PR #7, opened 2015-01-12
Adjust plugin testing

EarthmanT:
assert raises thing
move test dependencies to test-requirements.txt
rename utility to utils
rename "amazon" to ec2_client


ec2/instance.py
```diff
@@ -47,6 +47,7 @@
 def create(**kwargs):
     """ Creates an EC2 instance from an (AMI) image_id and an instance_type.
     """
+    ec2client = connection.EC2Client().ec2client()
```
nir0s: ec2_client = connection.EC2Client().connect()

PR #14, opened 2015-02-19
Example

EarthmanT:
Bootstrap manager works and nodecellar deployment from a manager works.
Before releasing to the work, system tests needed. also, see hackpad for items I would do to it:

https://cloudify.hackpad.com/Amazon-Plugin-fwVtKiuqFNt


ec2/configure.py
```diff
@@ -0,0 +1,54 @@
...
+#    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    * See the License for the specific language governing permissions and
+#    * limitations under the License.
+
+# Built-in Imports:
+import os
+import tempfile
+
+# Third Party Imports
+from boto import config
```
nir0s: `import boto` instead.

ec2/configure.py
```diff
@@ -0,0 +1,54 @@
...
+                     self._get_aws_credentials_name(),
+                     self._get_aws_access_key_id(),
+                     self._get_aws_secret_access_key())
+        return config
+
+    def _get_aws_credentials_name(self):
+        return config.get_value('Credentials', '__name__')
+
+    def _get_aws_access_key_id(self):
+        return config.get('Credentials', 'aws_access_key_id')
```
nir0s: boto.config.get(...)

ec2/configure.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def get_temp_file(self):
+        temp_config = tempfile.mktemp()
+        config = self._get_config()
+        with open(temp_config, 'w') as temp_config_file:
+            temp_config_file.write(config)
+        return temp_config
+
+    def _get_config(self):
+        config = '[{}]\n' \
```
nir0s: python 2.6 doesn't support empty format strings.
{0}...

ec2/configure.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def _get_aws_access_key_id(self):
+        return config.get('Credentials', 'aws_access_key_id')
+
+    def _get_aws_secret_access_key(self):
+        return config.get('Credentials', 'aws_secret_access_key')
+
+    def _get_path_to_boto_configuration(self):
+        home = os.path.expanduser('~')
+        path_to_boto_configuration = os.path.join(home, '.boto')
```
nir0s: It should be boldly and red-ly documented that the boto config file is always taken from the default locations and put at ~/.boto

ec2/tests/test_ec2_workflow.py
```diff
@@ -0,0 +1,142 @@
...
+        """ Tests the install workflow using the built in
+            workflows.
+        """
+        path = os.path.expanduser('~/.ssh')
+        file = os.path.join(path, '{0}{1}'.format('test_key', '.pem'))
+        if os.path.exists(file):
+            os.remove(file)
+
+        # execute install workflow
+        self.env.execute('install', task_retries=0)
```
nir0s: Ask Idan if there should be some kind of assertion here. This might prevent bugs within the code in case we return even though the workflow itself failed.

ec2/utils.py
```diff
@@ -13,61 +13,200 @@
...
+            parameters[key] = ctx.node.properties['parameters'][key]
+        elif key is 'image_id' or key is 'instance_type':
+            parameters[key] = ctx.node.properties[key]
+        else:
+            del(parameters[key])
+
+    return parameters
+
+
+def validate_node_property(key, ctx):
```
nir0s: Why do we need to validate that a property exists within ctx? We can just try and use it from within the task.

ec2/utils.py
```diff
@@ -13,61 +13,200 @@
...
-           before we stop checking.
-    :param check_interval: How long to wait between checks.
-    :return: bool (True the desired state was reached, False, it was not.)
+    if key not in ctx.node.properties.keys() \
+            or ctx.node.properties.get(key) is None:
+        raise NonRecoverableError('{0} is a required input .'
+                                  'Unable to create.'.format(key))
+
+
+def save_key_pair(key_pair_object, ctx):
```
nir0s: maybe all keypair related stuff should be under the key_pairs file?

ec2/utils.py
```diff
@@ -13,61 +13,200 @@
...
-                                                   check_interval))
+    instance_id = ctx.instance.runtime_properties['aws_resource_id']
+    instance = get_instance_from_id(instance_id, ctx=ctx)
+    if 'running' not in instance.update():
+        raise RecoverableError('Waiting for server to be running. '
+                               'Retrying...', retry_after=check_interval)
 
-    if check_interval < 1:
-        check_interval = 1
+    attribute = getattr(instance, attribute)
```
nir0s: maybe return getattr(instance, attribute)?

ec2/utils.py
```diff
@@ -13,61 +13,200 @@
...
+                        '{0}{1}'.format(
+                            ctx.node.properties['resource_id'],
+                            '.pem'))
+    if os.path.exists(file):
+        return True
+    else:
+        return False
+
+
+def get_instance_attribute(attribute, check_interval, ctx):
```
nir0s: `get_instance_attribute` should be a generic function for retrieving an attribute for an instance OBJECT, regardless of its state. We should check if the caller requires the instance to be running.. or so.

ec2/utils.py
```diff
@@ -13,61 +13,200 @@
...
+                                  check_interval, ctx=ctx)
 
+
+def get_public_ip_address(check_interval, ctx):
+    """ returns the public_ip_address variable for a given instance
+    """
+    return get_instance_attribute('ip_address', check_interval, ctx=ctx)
+
+
+def assign_runtime_properties_to_instance(retry_interval, ctx):
```
nir0s: probably not a util... should reside in instance..

ec2/utils.py
```diff
@@ -13,61 +13,200 @@
...
+            ctx.instance.runtime_properties['public_dns_name']))
+        ctx.logger.info('Public IP: {}.'.format(
+            ctx.instance.runtime_properties['public_ip_address']))
+        ctx.logger.info('Private DNS: {}.'.format(
+            ctx.instance.runtime_properties['private_dns_name']))
+        ctx.logger.info('Private IP (the ip): {}.'.format(
+            ctx.instance.runtime_properties['ip']))
+
+
+def get_instance_state(ctx):
```
nir0s: maybe merge with `get_instance_attribute`

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
-            raise NonRecoverableError('Timed out while attemting to get the '
-                                      'instance {0}. Timeout length: {1}.'
-                                      .format(attribute, timeout_length))
-    attribute = getattr(instance, attribute)
-    return attribute
+    kplist = get_target_aws_resource_id(relationship_type, ctx=ctx)
+    if kplist:
+        return kplist[0]
+    else:
+        return kplist
```
nir0s: return kplist[0] if kplist else kplist

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
 
     return instance
 
 
-def get_instance_attribute(instance, attribute, timeout_length):
-    """ given the related instance object
-        the given variable can be retrieved and returned
-    """
-    timeout = time.time() + timeout_length
+def get_attached_keypair_id(ctx):
```
nir0s: is this a util?

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
-                                  'public_dns_name', timeout_length)
 
+    ids = []
 
-def get_private_ip_address(instance, timeout_length):
-    """ returns the private_ip_address variable for a given instance
-    """
-    return get_instance_attribute(instance,
-                                  'private_ip_address', timeout_length)
+    if not hasattr(ctx.instance, 'relationships'):
```
nir0s: check whether an instance always has a relationships object even if no relationships are defined.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
 
+def get_security_group_from_id(group_id, ctx):
+    ec2_client = connection.EC2ConnectionClient().client()
+    ctx.logger.debug('Getting Security Group by ID: {0}'.format(group_id))
+
+    try:
+        group = ec2_client.get_all_security_groups(group_ids=group_id)
+    except boto.exception.EC2ResponseError as e:
+        group = get_security_group_from_name(group_id, ctx=ctx)
+        raise NonRecoverableError('Error. '
```
nir0s: if you actually managed to retrieve the group... let's let the user enjoy their mistake.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+                                  'Failed to group by id: '
+                                  'API returned: {0}.'
+                                  .format(str(e)))
 
-def validate_group(group, ctx):
-    ctx.logger.debug('Testing if group with identifier '
-                     ' {0} exists in this account.'.format(group))
-    groups = get_security_group_from_id(group, ctx)
+    if len(group) < 1:
```
nir0s: if not group

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+        group = get_security_group_from_name(group_id, ctx=ctx)
```
nir0s: this is probably redundant as you check this in line 287

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+    elif len(group) > 1:
```
nir0s: add documentation for why you did that please.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+        report_all_security_groups(ctx=ctx)
```
nir0s: can this really happen?

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
 
-    if groups is not None:
-        return True
-    else:
-        return False
+    return group
 
 
-def get_security_group_from_id(group, ctx):
+def get_security_group_from_name(group_name, ctx):
```
nir0s: maybe this should merged with get_security_group_from_id

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+                                  'Too many groups returned.')
 
+    return group
 
-def save_key_pair(key_pair_object, ctx):
-    """ Saves the key pair to the file specified in the blueprint
-    """
 
-    ctx.logger.debug('Attempting to save the key_pair_object.')
+def get_key_pair_by_id(ctx):
```
nir0s: We're creating connection clients all the time... maybe we should have the developer declare the client once and then send it to every function they want to execute.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+    ec2_client = connection.EC2ConnectionClient().client()
 
     try:
-        key_pair_object.save(ctx.node.properties['private_key_path'])
-    except OSError:
-        raise NonRecoverableError('Unable to save key pair to file: {0}.'
-                                  'OS Returned: {1}'.format(
-                                      ctx.node.properties['private_key_path'],
-                                      OSError))
+        key = ec2_client.get_key_pair(ctx.node.properties['resource_id'])
```
nir0s: maybe aws_resource_id...

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+            boto.exception.BotoServerError) as e:
+        raise NonRecoverableError('Error. '
+                                  'Failed to get key pair: '
+                                  'API returned: {0}.'
+                                  .format(str(e)))
+
+    return key.name
+
+
+def get_address_by_id(address_id, ctx):
```
nir0s: this should call get_address_object_by_id

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+                                  .format(str(e)))
+    except (boto.exception.BotoServerError) as e:
+        raise NonRecoverableError('Error. '
+                                  'Failed to get address by id: '
+                                  'API returned: {0}.'
+                                  .format(str(e)))
+    return address
+
+
+def report_all_instances(ctx):
```
nir0s: this could probably also be `get_all_instances` and return the list. might be useful.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+                 for instance in res.instances]
+
+    message = []
+    for instance in instances:
+        message.append('{}\n'.format(instance))
+
+    ctx.logger.info('Available instances: {0}'.format(message))
+
+
+def report_all_security_groups(ctx):
```
nir0s: same.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+                                  'API returned: {0}.'
+                                  .format(str(e)))
+    message = []
+    for group in groups:
+        message.append('{}\n'.format(group))
+
+    ctx.logger.info('Available groups: {0}'.format(message))
+
+
+def report_all_addresses(ctx):
```
nir0s: same old.

ec2/utils.py
```diff
@@ -76,120 +215,267 @@ def get_instance_state(instance, ctx):
...
+                                  'API returned: {0}.'
+                                  .format(str(e)))
+    message = []
+    for address in addresses:
+        message.append('{}\n'.format(address))
+
+    ctx.logger.info('Available addresses: {0}'.format(message))
+
+
+def validate_state(instance_id, state, timeout_length, check_interval, ctx):
```
nir0s: we should check with Idan why we need to this if we have `get_state`'

tox.ini
```diff
@@ -6,7 +6,7 @@ envlist=flake8,py27
 deps =
     -rdev-requirements.txt
     -rtest-requirements.txt
-commands=nosetests --with-cov --cov-report term-missing --cov ec2 ec2/tests
+commands=nosetests --nocapture --nologcapture --with-cov --cov-report term-missing --cov ec2 ec2/tests
```
nir0s: add -v

EarthmanT: done

ec2/utils.py
EarthmanT: Changed to get_all_addresses and if it fails it runs 'log_available_resources' and then raises non-recoverable

ec2/utils.py
EarthmanT: Changed to get_all_security_groups and if it fails it runs 'log_available_resources' and then raises non-recoverable

ec2/utils.py
EarthmanT: Changed to get_all_instances and if it fails it runs 'log_available_resources' and then raises non-recoverable

ec2/utils.py
EarthmanT: done

ec2/utils.py
EarthmanT: Moved to keypair.py and just pass the path from outside of the functon

ec2/utils.py
EarthmanT: refactored this whole function

ec2/utils.py
EarthmanT: re-factored entire functon

ec2/utils.py
EarthmanT: re-factored entire functon

ec2/utils.py
EarthmanT: I started doing what you suggested about keypair (return group[0] if group else group ) for everything

EarthmanT: not quite everything but close

ec2/utils.py
EarthmanT: yeah i preserved that functionlity using regex which is close to what amazon does

ec2/utils.py
EarthmanT: changed to use getattr(ctx.instance, 'relationships', []) so that if an empty list is returned from the value of relationships then it's skipped;

ec2/utils.py
EarthmanT: I LOVE IT

ec2/utils.py
EarthmanT: nah I don't think so.

ec2/utils.py
EarthmanT: done

ec2/utils.py
EarthmanT: moved

ec2/utils.py
EarthmanT: done

ec2/utils.py
EarthmanT: done

ec2/utils.py
EarthmanT: moved

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
     try:
-        address_object = ec2_client.allocate_address()
-    except (EC2ResponseError, BotoServerError) as e:
-        raise NonRecoverableError('Error. Failed to '
-                                  'provision Elastic IP. Error: {0}.'
-                                  .format(e))
+        address_object = ec2_client.allocate_address(domain=None)
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
+        raise NonRecoverableError('Failed to provision Elastic IP. Error: {0}.'
```
nir0s: please state why this is a NonRecoverableError (in the docs as well)

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
+        address_object.delete()
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error. Failed to '
                                   'delete Elastic IP. Error: {0}.'
                                   .format(e))
-
-    ctx.logger.info('Released Elastic IP {0}.'.format(elasticip))
+    except AttributeError:
+        ctx.logger.debug('Attribute error raised on address_object.release.')
```
nir0s: The error messages isn't clear...

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
 @operation
-def release(**kwargs):
+def release(**_):
     """ Releases an Elastic IP from the connected region in the AWS account.
     """
-    ec2_client = connection.EC2ConnectionClient().client()
     ctx.logger.info('Releasing an Elastic IP.')
 
-    elasticip = ctx.instance.runtime_properties['elasticip']
+    elasticip = ctx.instance.runtime_properties.get('aws_resource_id')
```
nir0s: `elasticip` might be None... we should probably check before continuing.

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
                                   .format(e))
-
-    ctx.logger.info('Released Elastic IP {0}.'.format(elasticip))
+    except AttributeError:
+        ctx.logger.debug('Attribute error raised on address_object.release.')
+        pass
+    finally:
+        elasticip = \
+            ctx.instance.runtime_properties.pop('aws_resource_id', None)
+        ctx.instance.runtime_properties.pop('allocation_id', None)
```
nir0s: pop -> del

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
     """
-    ec2_client = connection.EC2ConnectionClient().client()
     ctx.logger.info('Releasing an Elastic IP.')
 
-    elasticip = ctx.instance.runtime_properties['elasticip']
+    elasticip = ctx.instance.runtime_properties.get('aws_resource_id')
+
+    address_object = utils.get_address_object_by_id(elasticip, ctx=ctx)
 
     try:
```
nir0s: finally is probably not the correct action to perform here...
* If we managed to release the address, all is good.
* else if we didn't, we raise an NRE.
* else if it's a VPC IP, instead of passing, we should raise an NRE because the address could not be released.

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
-def associate(**kwargs):
+def associate(**_):
     """ Associates an Elastic IP with an EC2 Instance.
     """
     ec2_client = connection.EC2ConnectionClient().client()
 
-    elasticip = ctx.target.node.properties['elasticip']
-    instance_id = ctx.source.instance.runtime_properties['instance_id']
+    instance_id = ctx.source.instance.runtime_properties.get('aws_resource_id')
+    elasticip = ctx.target.instance.runtime_properties.get('aws_resource_id')
```
nir0s: for `instance_id` and `elasticip`, we need to verify that they're not None.`

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
 
 
 @operation
-def disassociate(**kwargs):
+def disassociate(**_):
     """ Disassociates an Elastic IP from an EC2 Instance.
     """
     ec2_client = connection.EC2ConnectionClient().client()
-    elasticip = ctx.target.node.properties['elasticip']
+    elasticip = ctx.target.instance.runtime_properties['aws_resource_id']
```
nir0s:
```python
try:
    elasticip = ctx.target.instance.runtime_properties['aws_resource_id']
except KeyError as ex:
    ...
```

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
     try:
         ec2_client.disassociate_address(public_ip=elasticip)
-    except (EC2ResponseError, BotoServerError) as e:
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error. Failed to detach '
                                   'Elastic IP, returned: {0}.'
                                   .format(e))
+    finally:
+        ctx.source.instance.runtime_properties.pop('public_ip_address')
```
nir0s: pop -> del

ec2/elasticip.py
```diff
@@ -14,109 +14,123 @@
...
 
     try:
         ec2_client.disassociate_address(public_ip=elasticip)
-    except (EC2ResponseError, BotoServerError) as e:
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error. Failed to detach '
                                   'Elastic IP, returned: {0}.'
                                   .format(e))
+    finally:
```
nir0s: finally is pretty much useless here..

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
-                                   RUN_INSTANCES_UNSUPPORTED)
-    arguments.update(args_to_merge)
+    if ctx.node.properties['use_external_resource']:
+        instance_id = ctx.node.properties['resource_id']
+        instance = utils.get_instance_from_id(instance_id, ctx=ctx)
+        ctx.instance.runtime_properties['aws_resource_id'] = instance.id
+        ctx.logger.info('Using existing resource: {0}.'.format(instance.id))
+        return
+
+    parameters = utils.get_parameters(ctx=ctx)
```
nir0s: maybe

```python
instance_parameters = utils.get_instance_parameters()
```

EarthmanT: made the change

ec2/elasticip.py
EarthmanT: ok

ec2/elasticip.py
EarthmanT: ok

ec2/elasticip.py
EarthmanT: ok

ec2/elasticip.py
EarthmanT: that was simple enough

ec2/elasticip.py
EarthmanT: Super clear now

ec2/elasticip.py
EarthmanT: love del

nir0s: Lol

ec2/elasticip.py
EarthmanT: opted for:
```
    if key not in dict:
       raise
```

ec2/elasticip.py
EarthmanT: It's not idempotent.
I would propose that we do a test to make sure that it was provisioned (which would be slow) rather than allow it to be retried.

ec2/elasticip.py
```diff
@@ -14,109 +14,158 @@
...
 
 @operation
-def disassociate(**kwargs):
+def disassociate(**_):
     """ Disassociates an Elastic IP from an EC2 Instance.
     """
     ec2_client = connection.EC2ConnectionClient().client()
-    elasticip = ctx.target.node.properties['elasticip']
+
+    if 'aws_resource_id' not in ctx.target.instance.runtime_properties:
```
nir0s: Maybe this should be done everywhere (where relevant!)

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
+def stop(retry_interval, **_):
     """ Stops an existing EC2 instance.
         If already stopped, this does nothing.
+        Requires:
+            ctx.instance.runtime_properties['aws_resource_id']
     """
     ec2_client = connection.EC2ConnectionClient().client()
 
-    instance_id = ctx.instance.runtime_properties['instance_id']
+    instance_id = ctx.instance.runtime_properties.get('aws_resource_id')
```
nir0s: if something in something..  raise

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
 
     ctx.logger.info('Stopping EC2 Instance.')
     ctx.logger.debug('Attempting to stop EC2 Instance.'
                      '(Instance id: {0}.)'.format(instance_id))
 
     try:
-        instances = ec2_client.stop_instances(instance_id)
-    except (EC2ResponseError, BotoServerError) as e:
+        ec2_client.stop_instances(instance_id)
```
nir0s: maybe we should verify throughout the plugin that when we try to perform actions on actioned resources, we don't get an exception.. but rather boto's implementation is idempotent for all actions....

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error. Failed to stop EC2 Instance: '
-                                  'API returned: {0}.'.format(e))
-
-    if utils.validate_instance_id(instance_id, ctx=ctx):
-        utils.validate_state(instances[0], INSTANCE_STOPPED,
-                             STOP_TIMEOUT, CHECK_INTERVAL, ctx=ctx)
+                                  'API returned: {0}.'.format(str(e)))
+    finally:
+        ctx.instance.runtime_properties.pop('private_dns_name')
```
nir0s: del del del del

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error. Failed to stop EC2 Instance: '
-                                  'API returned: {0}.'.format(e))
-
-    if utils.validate_instance_id(instance_id, ctx=ctx):
-        utils.validate_state(instances[0], INSTANCE_STOPPED,
-                             STOP_TIMEOUT, CHECK_INTERVAL, ctx=ctx)
+                                  'API returned: {0}.'.format(str(e)))
+    finally:
```
nir0s: the popping below should go under a `check_state` section and be del instead.

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
 @operation
-def terminate(**kwargs):
+def terminate(retry_interval, **_):
     """ Terminates an existing EC2 instance.
         If already terminated, this does nothing.
     """
     ec2_client = connection.EC2ConnectionClient().client()
 
-    instance_id = ctx.instance.runtime_properties['instance_id']
+    instance_id = ctx.instance.runtime_properties.get('aws_resource_id')
```
nir0s: if `x` not in `y'... raise

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
-        instances = ec2_client.terminate_instances(instance_id)
-    except (EC2ResponseError, BotoServerError) as e:
+        ec2_client.terminate_instances(instance_id)
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error. Failed to terminate '
                                   'EC2 Instance: API returned: {0}.'
-                                  .format(e))
+                                  .format(str(e)))
+    finally:
```
nir0s: only `aws_resource_id` is relevant here. Also, shouldn't be in finally but rather as a consequence of the validation.

ec2/elasticip.py
```diff
@@ -14,109 +14,158 @@
...
-    except (EC2ResponseError, BotoServerError) as e:
-        raise NonRecoverableError('Error. Failed to validate '
-                                  'Elastic IP, returned: {0}.'
-                                  .format(e))
 
-    ctx.logger.info('Verified that Elastic IP {0} was created.'.format(
-        elasticip))
-    return True
+@operation
+def creation_validation(**_):
```
nir0s: Ask Idan why we need creation_validation.. at all!

nir0s: Don't let him convince you easily!

nir0s: he's really good at it!

ec2/instance.py
```diff
@@ -14,165 +14,203 @@
...
+
+        ctx.instance.runtime_properties['private_dns_name'] = \
+            utils.get_private_dns_name(ctx=ctx)
+        ctx.instance.runtime_properties['public_dns_name'] = \
+            utils.get_public_dns_name(ctx=ctx)
+        ctx.instance.runtime_properties['public_ip_address'] = \
+            utils.get_public_ip_address(ctx=ctx)
+        ctx.instance.runtime_properties['ip'] = \
+            utils.get_private_ip_address(ctx=ctx)
+        ctx.logger.info('Public DNS: {}.'.format(
```
nir0s: {} -> {0}.. throughout the entire package.

ec2/keypair.py
```diff
@@ -30,21 +31,45 @@
...
+
     ec2_client = connection.EC2ConnectionClient().client()
-    key_pair_name = ctx.node.properties['name']
+
+    if ctx.node.properties['use_external_resource']:
+        key_pair_id = ctx.node.properties['resource_id']
+        key_pair = utils.get_key_pair_by_id(key_pair_id)
+        ctx.instance.runtime_properties['aws_resource_id'] = key_pair.name
+        key_path = get_key_file_path(ctx=ctx)
+        ctx.logger.info(key_path)
```
nir0s: maybe a bit more verbose.

ec2/keypair.py
```diff
@@ -30,21 +31,45 @@
...
+        if not search_for_key_file(key_path):
+            raise NonRecoverableError('use_external_resource was specified, '
+                                      'and a name given, but the key pair was '
+                                      'not located on the filesystem.')
+        return
+
+    key_pair_name = ctx.node.properties['resource_id']
+
     ctx.logger.info('Creating key pair.')
```
nir0s: this should probably be `debug` since the user doesn't care that you're currently creating a keypair.. they care that you created it.

we should check that `debug` is viewable in the logs..

ec2/keypair.py
```diff
@@ -30,21 +31,45 @@
...
 
     try:
         kp = ec2_client.create_key_pair(key_pair_name)
-    except (EC2ResponseError, BotoServerError, BotoClientError) as e:
-        raise NonRecoverableError('Key pair not created. '
-                                  'API returned: {0}'.format(e))
+    except (boto.exception.EC2ResponseError,
```
nir0s: in this case.. maybe it didn't fail because of a duplicate key.. but because it just failed to create (maybe.. too many keypairs in account).

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
         when this lifecycle operation is called.
     """
     ec2_client = connection.EC2ConnectionClient().client()
-    key_pair_name = ctx.instance.runtime_properties['key_pair_name']
+    key_pair_name = ctx.instance.runtime_properties['aws_resource_id']
```
nir0s: if `...` is not in ```...

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
 
     try:
         ec2_client.delete_key_pair(key_pair_name)
-    except (EC2ResponseError, BotoServerError) as e:
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError('Error response on key pair delete. '
-                                  'API returned: {0}'.format(e))
+                                  'API returned: {0}'.format(str(e)))
+    finally:
```
nir0s: first del.. and verify that deletion was successful before removing from ctx.

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+        ctx.instance.runtime_properties.pop('aws_resource_id', None)
+        ctx.instance.runtime_properties.pop('key_path', None)
+        ctx.logger.debug('Attempted to delete key pair from account.')
 
+    delete_key_file(ctx=ctx)
```
nir0s: only if it was actually deleted from the account.. delete from local.

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+    """ Saves the key pair to the file specified in the blueprint. """
+
+    ctx.logger.debug('Attempting to save the key_pair_object.')
 
     try:
-        ec2_client.get_key_pair(key_pair_name)
-    except (EC2ResponseError, BotoServerError) as e:
-        raise NonRecoverableError('Unable to validate that Key Pair exists. '
-                                  'API returned: {0}'.format(e))
+        key_pair_object.save(ctx.node.properties['private_key_path'])
```
nir0s: if `private_key_path` in ...

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+    except (boto.exception.BotoClientError,
+            OSError) as e:
+        raise NonRecoverableError(str(e))
+
+    key_path = get_key_file_path(ctx=ctx)
+
+    os.chmod(key_path, 0600)
```
nir0s: check that you have permissions to perform this.

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+
+
+def get_key_file_path(ctx):
+    """The key_path is an attribute that gives the full path to the key file.
+    This function creates the path as a string for use by various functions in
+    this module. It doesn't verify whether the path points to anything.
+    """
+
+    path = os.path.expanduser(ctx.node.properties['private_key_path'])
```
nir0s: check if in.

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+    key_path = os.path.join(path,
+                            '{0}{1}'.format(
+                                ctx.node.properties['resource_id'],
```
nir0s: check if `aws_resource_id` has already been set in the workflow when this function runs.

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+
+def get_key_file_path(ctx):
+    """The key_path is an attribute that gives the full path to the key file.
+    This function creates the path as a string for use by various functions in
+    this module. It doesn't verify whether the path points to anything.
+    """
+
+    path = os.path.expanduser(ctx.node.properties['private_key_path'])
+    key_path = os.path.join(path,
+                            '{0}{1}'.format(
```
nir0s: {0}.pem

ec2/keypair.py
```diff
@@ -53,29 +78,86 @@ def delete(**kwargs):
...
+
+
+def search_for_key_file(key_path):
+    """ Indicates whether the file exists locally. """
 
-    ctx.logger.info('Validated key pair.')
+    if os.path.exists(key_path):
+        return True
+    else:
+        return False
```
nir0s: return True if os.path.exists(key_path) else False

ec2/securitygroup.py
```diff
@@ -26,211 +25,128 @@
...
+            rules: The rules which we want to create for the group.
+        ctx.instance.runtime_properties:
+            aws_resource_id: This is the security group ID assigned
+              by Amazon when the group is created.
     """
     ec2_client = connection.EC2ConnectionClient().client()
 
-    name = ctx.node.properties['name']
-    description = ctx.node.properties['description']
+    if ctx.node.properties.get('use_external_resource', False) is True:
```
nir0s: fix.

ec2/securitygroup.py
```diff
@@ -26,211 +25,128 @@
...
+        group_id = ctx.node.properties.get('resource_id')
```
nir0s: fix.,

ec2/securitygroup.py
```diff
@@ -26,211 +25,128 @@
...
+        group = utils.get_security_group_from_id(group_id, ctx=ctx)
+        if not group:
```
nir0s: check if `get_security_group_from_id` raises the error.

ec2/instance.py
EarthmanT: done

ec2/instance.py
EarthmanT: Start, stop, terminate are all idempotent.

ec2/instance.py
EarthmanT: checked couldn't find any more

ec2/instance.py
EarthmanT: done

ec2/instance.py
EarthmanT: done

ec2/instance.py
EarthmanT: done

ec2/keypair.py
EarthmanT: more verbose

ec2/keypair.py
EarthmanT: done

ec2/keypair.py
EarthmanT: ok

ec2/keypair.py
EarthmanT: great

ec2/keypair.py
EarthmanT: checked

ec2/keypair.py
EarthmanT: done

ec2/keypair.py
EarthmanT: checked

ec2/keypair.py
EarthmanT: added

ec2/keypair.py
EarthmanT: added

ec2/keypair.py
EarthmanT: added

ec2/keypair.py
EarthmanT: done

ec2/elasticip.py
EarthmanT: done

ec2/elasticip.py
```diff
@@ -14,109 +14,168 @@
...
+            ctx.node.properties['resource_id'], ctx=ctx)
+        if not address:
+            raise NonRecoverableError(
+                'External elasticip was indicated, but the given '
+                'elasticip does not exist.')
+        ctx.instance.runtime_properties['aws_resource_id'] = address
+        ctx.logger.info('Using existing resource: {0}'.format(address))
+        return
+
+    ctx.logger.debug('Attepting to allocate elasticip.')
```
EarthmanT: atteMpting

ec2/elasticip.py
```diff
@@ -14,109 +14,168 @@
...
-        ec2_client.release_address(public_ip=elasticip)
-    except (EC2ResponseError, BotoServerError) as e:
-        raise NonRecoverableError('Error. Failed to '
-                                  'delete Elastic IP. Error: {0}.'
-                                  .format(e))
+    if 'aws_resource_id' not in ctx.instance.runtime_properties:
+        raise NonRecoverableError(
+            'Unable to release elasticip, because aws_resource_id is not set.')
+
+    elasticip = ctx.instance.runtime_properties['aws_resource_id']
```
EarthmanT: in __init__ add 'resource_title' or something = 'aws_resource_id'

ec2/elasticip.py
```diff
@@ -14,109 +14,168 @@
...
+        elasticip = \
+            ctx.instance.runtime_properties.pop('aws_resource_id', None)
+        if 'allocation_id' in ctx.instance.runtime_properties:
+            del(ctx.instance.runtime_properties['allocation_id'])
+        ctx.logger.info(
+            'Released elastic ip {0}. '
+            'Removed runtime properties.'.format(elasticip))
+    else:
+        raise RecoverableError(
+            'Elastic IP not provisioned. Retrying...',
```
EarthmanT: not released

ec2/elasticip.py
```diff
@@ -14,109 +14,168 @@
...
-def disassociate(**kwargs):
+def disassociate(**_):
     """ Disassociates an Elastic IP from an EC2 Instance.
     """
     ec2_client = connection.EC2ConnectionClient().client()
-    elasticip = ctx.target.node.properties['elasticip']
+
+    if 'aws_resource_id' not in ctx.target.instance.runtime_properties:
+        raise NonRecoverableError(
+            'Failed to disossiate elastic ip, '
```
EarthmanT: discuss with ran if external_resource should have relationship operations performed on them or not.

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
-
-    arguments = {}
-    for pair in user_supplied.items():
-        arguments['{0}'.format(pair[0])] = pair[1]
-    for pair in unsupported.items():
-        arguments['{0}'.format(pair[0])] = pair[1]
-    return arguments
+@operation
+def creation_validation(**_):
+    """ This checks that all user supplied info is valid """
```
EarthmanT: ad validation for flavor and image id

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
-                     'Image id: {0}. Instance type: {1}.'
-                     .format(arguments['image_id'],
-                             arguments['instance_type']))
-    ctx.logger.debug('Sending these API parameters: {0}.'
-                     .format(arguments))
+    if ctx.node.properties['use_external_resource']:
+        instance_id = ctx.node.properties['resource_id']
+        instance = utils.get_instance_from_id(instance_id, ctx=ctx)
+        ctx.instance.runtime_properties['aws_resource_id'] = instance.id
+        ctx.logger.info('Using existing instance: {0}.'.format(instance.id))
```
EarthmanT: raise error if instance is none

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
+    ctx.logger.debug('Attempting to start instance: {0}.)'.format(instance_id))
 
-    if utils.validate_instance_id(instance_id, ctx=ctx):
-        utils.validate_state(instances[0], INSTANCE_RUNNING,
-                             START_TIMEOUT, CHECK_INTERVAL, ctx=ctx)
+    try:
+        ec2_client.start_instances(instance_id)
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
+        if 'does not exist' in e:
```
EarthmanT: found out if theres a better solution

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
+        raise NonRecoverableError('{0}'.format(str(e)))
 
     instance_id = reservation.instances[0].id
-    ctx.instance.runtime_properties['instance_id'] = instance_id
-
-    if utils.validate_instance_id(reservation.instances[0].id, ctx=ctx):
-        utils.validate_state(reservation.instances[0], INSTANCE_RUNNING,
-                             CREATION_TIMEOUT, CHECK_INTERVAL, ctx=ctx)
+    ctx.instance.runtime_properties['aws_resource_id'] = instance_id
+    ctx.logger.info('Created instance: {0}.'.format(instance_id))
```
EarthmanT: check that id exists rather than doing the 'does not exist' thing in the start call down below

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
+
+    ctx.logger.debug('Attempted to start instance {0}.'.format(instance_id))
+
+    if utils.get_instance_state(ctx=ctx) == 16:
+        assign_runtime_properties_to_instance(ctx=ctx)
+        ctx.logger.info('Instance {0} is running.'.format(instance_id))
+    else:
+        raise RecoverableError(
+            'Waiting for server to be running. Retrying...',
+            retry_after=retry_interval)
```
EarthmanT: change return ctx.operation.retry()

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
 
 @operation
-def stop(**kwargs):
+def stop(retry_interval, **_):
     """ Stops an existing EC2 instance.
         If already stopped, this does nothing.
+        Requires:
+            ctx.instance.runtime_properties['aws_resource_id']
     """
     ec2_client = connection.EC2ConnectionClient().client()
```
EarthmanT: check with ran about use_external_resource "stopping" yea or nay

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
-    if utils.validate_instance_id(instance_id, ctx=ctx):
-        utils.validate_state(instances[0], INSTANCE_STOPPED,
-                             STOP_TIMEOUT, CHECK_INTERVAL, ctx=ctx)
+    if utils.get_instance_state(ctx=ctx) == 80:
+        ctx.logger.info('Stopped instance {0}.'.format(instance_id))
+        unassign_runtime_properties(ctx=ctx)
+    else:
+        raise RecoverableError(
+            'Waiting for server to stop. Retrying...',
+            retry_after=retry_interval)
```
EarthmanT: return ctx.operation.retry

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
+        raise NonRecoverableError('{0}'.format(str(e)))
+
+    ctx.logger.debug('Attempted to stop instance {0}.'.format(instance_id))
 
-    if utils.validate_instance_id(instance_id, ctx=ctx):
-        utils.validate_state(instances[0], INSTANCE_STOPPED,
-                             STOP_TIMEOUT, CHECK_INTERVAL, ctx=ctx)
+    if utils.get_instance_state(ctx=ctx) == 80:
```
EarthmanT: from ec2 import constants
constants.STATE_STOPPED

ec2/instance.py
```diff
@@ -14,165 +14,208 @@
...
-    ctx.logger.debug('Attempting to terminate EC2 Instance.'
-                     '(Instance id: {0}.)'.format(instance_id))
-
-    try:
-        instances = ec2_client.terminate_instances(instance_id)
-    except (EC2ResponseError, BotoServerError) as e:
-        raise NonRecoverableError('Error. Failed to terminate '
-                                  'EC2 Instance: API returned: {0}.'
-                                  .format(e))
+    if 'aws_resource_id' not in ctx.instance.runtime_properties:
```
EarthmanT: no use_external_resource

ec2/elasticip.py
EarthmanT: done

ec2/elasticip.py
EarthmanT: constants.EXTERNAL_RESOURCE_ID

ec2/elasticip.py
EarthmanT: ok

ec2/instance.py
EarthmanT: ok

ec2/elasticip.py
EarthmanT: discussed

PR #16, opened 2015-03-24
updating the parameters function

EarthmanT:


ec2/instance.py
```diff
@@ -389,44 +389,17 @@ def _get_instance_parameters(node_properties, ctx_instance, ctx_logger):
...
-            else:
-                parameters[key] = \
-                    _get_instance_keypair(ctx_instance, ctx_logger)
-        elif key in node_parameter_keys:
-            parameters[key] = node_properties['parameters'][key]
-        elif key is 'image_id' or key is 'instance_type':
-            parameters[key] = node_properties[key]
-        else:
-            del(parameters[key])
+    parameters = {}
```
idanmo: You can have initialization and keys setting in the same line...
parameters = {
  'key': 'value'
}

PR #22, opened 2015-03-26
Support provider context

EarthmanT:
What did we do here?
- I had to change each of the references to the plugin.yaml to use the branch plugin.yaml (or archive) because the tests need to be on the right branch.
- I changed the bootstrap configure script to put the "provider context" in the manager home directory.
- I updated the name of the task in that configure script that is used by fabric in the manager configure operation
- I updated the tests for the new "provider context" implementation.
- I updated the keypair.py creation_validation function to work with the expected behavior of get_keypair_by_id in certain circumstances, which was changed in the last PR.
- I created an implementation for "provider_context" which is not really a context, but a dict that goes on the manager.
- I added a fix for the issue where amazon doesn't update the API or return a good instance_id when you run instances.


ec2/constants.py
```diff
@@ -47,3 +47,6 @@
 
 # keypair module constants
 KEYPAIR_REQUIRED_PROPERTIES = ['private_key_path']
+
+# Config
+AWS_CONFIG_PATH = '~/.aws_config.json'
```
EarthmanT: dont hide

ec2/instance.py
```diff
@@ -206,19 +206,48 @@ def _unassign_runtime_properties(runtime_properties, ctx_instance):
...
+                    ctx.instance.runtime_properties['reservation_id']
+            })
+    except (boto.exception.EC2ResponseError,
+            boto.exception.BotoServerError) as e:
         raise NonRecoverableError(
-            'Instance failed for an unknown reason. Node ID: {0}.'
-            .format(ctx.instance.id))
-    else:
-        return ctx.instance.runtime_properties[constants.EXTERNAL_RESOURCE_ID]
+            '{1}'.format(str(e)))
```
EarthmanT: fix that

ec2/instance.py
```diff
@@ -358,9 +390,14 @@ def _get_instance_attribute(attribute):
     instance_object = _get_instance_from_id(instance_id)
 
     if instance_object is None:
-        raise NonRecoverableError(
-            'Unable to get instance attibute {0}, because no instance with id '
-            '{1} exists in this account.'.format(attribute, instance_id))
+        ec2_client = connection.EC2ConnectionClient().client()
+        instances = _get_instance_id_from_reservation_id(ec2_client)
```
EarthmanT: check if this can go in get instance from id

EarthmanT: create a default flow

ec2/instance.py
```diff
@@ -358,9 +390,14 @@ def _get_instance_attribute(attribute):
     instance_object = _get_instance_from_id(instance_id)
 
     if instance_object is None:
```
EarthmanT: and not external resource
....
else
raise NonRecoverableError because external resource doesn't have reservation id

ec2/instance.py
```diff
@@ -383,16 +420,23 @@ def _get_instance_parameters():
...
     """
 
+    provider_variables = utils.get_provider_variables()
+
     attached_group_ids = \
         utils.get_target_external_resource_ids(
             constants.INSTANCE_SECURITY_GROUP_RELATIONSHIP,
             ctx.instance)
 
+    if 'agents_security_group' in provider_variables and \
```
EarthmanT:
```
if provider_variables.get('agents_security_group'):
  ...
```

ec2/instance.py
```diff
@@ -409,7 +453,10 @@ def _get_instance_keypair():
         utils.get_target_external_resource_ids(
             constants.INSTANCE_KEYPAIR_RELATIONSHIP, ctx.instance)
 
-    if len(list_of_keypairs) > 1:
+    if not list_of_keypairs and 'agents_keypair' in provider_variables \
```
EarthmanT: same `get` as before

ec2/instance.py
```diff
@@ -409,7 +453,10 @@ def _get_instance_keypair():
...
+            and provider_variables['agents_keypair'] is not None:
+        list_of_keypairs.append(provider_variables['agents_keypair'])
+    elif len(list_of_keypairs) > 1:
         raise NonRecoverableError(
             'Only one keypair may be attached to an instance.')
```
EarthmanT: `else None`

ec2/keypair.py
```diff
@@ -38,14 +38,14 @@ def creation_validation(**_):
...
         if not key_file_in_filesystem:
             raise NonRecoverableError(
                 'External resource, but the key file does not exist locally.')
-        if not key_pair_in_account:
+        try:
+            _get_key_pair_by_id(ctx.node.properties['resource_id'])
+        except NonRecoverableError:
             raise NonRecoverableError(
                 'External resource, '
                 'but the key pair does not exist in the account.')
```
EarthmanT: as e
and add the original message

ec2/keypair.py
```diff
@@ -55,7 +55,11 @@ def creation_validation(**_):
             raise NonRecoverableError(
                 'Not external resource, '
                 'but the key file exists locally.')
-        if key_pair_in_account:
+        try:
```
EarthmanT: Look up -> should be `else`

ec2/constants.py
EarthmanT: First, make this AWS_DEFAULT_CONFIG_PATH.

Then, fabric call in bootstrap will have an optional non-default path.

And utils will see if environment variable exists and read from it or fall back to the default path.

And we will rereview this

ec2/utils.py
```diff
@@ -160,3 +161,70 @@ def get_resource_id():
...
+    }
+
+    ctx.logger.info(provider_context)
+
+    return provider_context
+
+
+def _get_variable(variable_name):
+
+    if variable_name in os.environ:
```
EarthmanT: no no no

system_tests/resources/scripts/configure.py
```diff
@@ -0,0 +1,45 @@
...
+
+def set_provider_context(agents_security_group, agents_keypair):
+    temp_config = tempfile.mktemp()
+
+    provider_context_json = {
+        "agents_keypair": agents_keypair,
+        "agents_security_group": agents_security_group
+    }
+
+    os.environ['agents_keypair'] = agents_keypair
```
EarthmanT: no no no

system_tests/resources/scripts/configure.py
```diff
@@ -0,0 +1,45 @@
...
+
+import os
+import json
+import tempfile
+import fabric.api
+from ec2 import configure
+from ec2 import constants
+
+def configure_manager(config_path, agents_security_group, agents_keypair):
+    upload_credentials(config_path)
```
EarthmanT: make internal private

ec2/instance.py
```diff
@@ -357,10 +388,20 @@ def _get_instance_attribute(attribute):
         ctx.instance.runtime_properties[constants.EXTERNAL_RESOURCE_ID]
     instance_object = _get_instance_from_id(instance_id)
 
-    if instance_object is None:
+    if not instance_object \
```
dankilman: nested condition
`if not instance_object:`

plugin.yaml
```diff
@@ -6,7 +6,7 @@ tosca_definitions_version: cloudify_dsl_1_1
 plugins:
   aws:
     executor: central_deployment_agent
-    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/master.zip
+    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/support_provider_context.zip
```
dankilman: change to master when all is done

EarthmanT: did

system_tests/manager/ec2_nodecellar_test.py
```diff
@@ -17,7 +17,7 @@
 from cosmo_tester.test_suites.test_blueprints import nodecellar_test
 
 EXTERNAL_RESOURCE_ID = 'aws_resource_id'
-CLOUDIFY_EC2_NODECELLAR = '/home/trammell/Environments/cfy3v2m6/' \
+CLOUDIFY_EC2_NODECELLAR = '~/Environments/' \
     'cloudify-nodecellar-example'
```
EarthmanT: did

PR #45, opened 2015-04-15
TFY-3 Remove DSL Parser Type Asserts Replace with Asserts Against EC2

EarthmanT:


ec2/tests/test_ec2_workflow.py
```diff
@@ -80,12 +82,12 @@ def test_uninstall_workflow(self):
             workflows.
         """
 
-        # execute install workflow
-        self.env.execute('uninstall', task_retries=0)
         path = os.path.expanduser('~/.ssh')
```
EarthmanT: tfy (tempfile)

ec2/tests/test_ec2_workflow.py
```diff
@@ -133,11 +135,14 @@ def test_uninstall_workflow(self):
             workflows.
         """
 
+        def clean_up_keys():
+            path = os.path.expanduser('~/.ssh')
```
EarthmanT: same

system_tests/local/test_plugin.py
```diff
@@ -37,7 +33,12 @@
 
 class TestWorkflowClean(EC2LocalTestUtils):
 
+    def tearDown(self):
+        super(TestWorkflowClean, self).tearDown()
+        self.localenv.execute('uninstall', task_retries=10)
```
EarthmanT: assert that after the uninstall these operations are still valid. (instance_id or security group is not there anymore)
Then it's no longer part of the teardown. it has some assertions

EarthmanT: i did

PR #67, opened 2015-10-04
delete ec2_region_name property when value is ''

ChenRoth:
fix bug where ec2_region_name property is not deleted if it's '', resulting in an unexpected input error when instantiating an EC2Connection
(i haven't added any tests)


nir0s:
If i'm not mistaken, there should be a test for that (using moto or somn) which should now be modified, if not, @EarthmanT, what do you think?

PR #72, opened 2015-11-09
Following Code Review

EarthmanT:
Added more validation and updated tests

### DIFF
--- a/ec2/ebs.py
+++ b/ec2/ebs.py
@@ -14,7 +14,6 @@
 #    * limitations under the License.
 
 # Built in Imports
-import re
 import datetime
 
 # Third-party Imports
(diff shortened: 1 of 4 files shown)


ec2/ebs.py
```diff
@@ -67,16 +64,15 @@ def create(args, **_):
 
     ctx.logger.debug('Creating EBS volume')
 
-    if not args:
-        args = dict(
-            size=ctx.node.properties['size'],
-            zone=ctx.node.properties['zone']
-        )
+    create_volume_args = dict(
```
m1keil: I wonder, any particular reason you use dict()?

Why not 

```
create_volume_args = {
   size: ...,
   zone: ....,
} 
```
?

ec2/ebs.py
```diff
@@ -338,15 +356,25 @@ def _get_volumes_from_id(volume_id):
...
+    :param status: a string ('available', 'creating', 'in-use')
+    :param volume: a volume object
+    :return: boolean
+    """
+
+    if status not in volume.update():
+        return False
+
+    return True
+
```
m1keil:
```
return status is volume.update()
```

PR #100, opened 2015-12-29
Cfy 4619 add region

EarthmanT:


EarthmanT:
cherry picked 7df17c1 and pushed

PR #101, opened 2015-12-29
Cfy 4620 allow route override

EarthmanT:


EarthmanT:
cherry picked the 16fea24 to avoid the other stuff

PR #105, opened 2015-12-30
added checks for default

EarthmanT:
added a check for security group vpc and subnet to check if they are default and to avoid if they are


system_tests/ec2_handler.py
```diff
@@ -359,12 +363,20 @@ def _elbs(self, elb_client):
...
-                for vpc in vpc_client.get_all_vpcs()]
+        vpcs = []
+        for vpc in vpc_client.get_all_vpcs():
+            if vpc.is_default:
+                continue
+            vpcs.append(vpc.id)
+        return vpcs
 
     def _subnets(self, vpc_client):
-        return [(subnet.id, subnet.id)
```
EarthmanT: [(subnet.id, subnet.id) for subnet in vpc_client.get_all_subnets() if not subnet.defaultForAz]

PR #117, opened 2016-01-07
CFY-4099 added windows password retrieval

nirbir:


ec2/passwd.py
```diff
@@ -0,0 +1,70 @@
...
+
+    decrypted_data = _long_to_bytes(plaintext)
+    unpadded_data = _pkcs1_unpad(decrypted_data)
+
+    return unpadded_data
+
+
+def get_windows_passwd(private_key_path, password_data):
+
+    key_file = open(private_key_path)
```
nirbir: close file
use with

PR #144, opened 2016-06-14
modify instance attribute to delete EBS volumes upon termination

Jessiebr:


EarthmanT:
We do not want to specify this in the attach operation.

PR #151, opened 2016-07-05
Cfy 5441 restructure aws plugin key pair class

Jessiebr:


cloudify_aws/ec2/keypair.py
```diff
@@ -0,0 +1,263 @@
...
+
+        return True
+
+    def created(self, args=None):
+
+        ctx.logger.info(
+                'Attempting to create keypair {0}.'
+                .format(self.cloudify_node_instance_id))
+
+        if self._create_external_keypair() or self.create(args):
```
EarthmanT: this should always be 
```python
    if self. self.use_external_resource_naively()...
```

If you need to change the implementation override the existing function

cloudify_aws/ec2/keypair.py
```diff
@@ -0,0 +1,263 @@
...
+        self._save_key_pair(kp)
+
+        ctx.instance.runtime_properties[constants.AWS_TYPE_PROPERTY] = \
+            constants.KEYPAIR['AWS_RESOURCE_TYPE']
+
+        self.resource_id = kp.name
+
+        return True
+
+    def created(self, args=None):
```
EarthmanT: Shouldn't need to override this...you override self.use_external_resource_naively

cloudify_aws/ec2/keypair.py
```diff
@@ -0,0 +1,263 @@
...
+    def _search_for_key_file(self, path_to_key_file):
+        """ Checks if the key_path exists in the local filesystem.
+
+        :param key_path: The path to the key pair file.
+        :return boolean if key_path exists (True) or not.
+        """
+
+        return True if os.path.exists(path_to_key_file) else False
+
+    def _get_key_pair_by_id(self, key_pair_id):
```
EarthmanT: you don't want this function, you want to use the built-in function get_all_matching

cloudify_aws/ec2/keypair.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if self._search_for_key_file(key_path):
+            try:
+                os.remove(key_path)
+            except OSError as e:
+                raise NonRecoverableError(
+                        'Unable to delete key pair: {0}.'
+                        .format(str(e)))
+
+    def _create_external_keypair(self):
```
EarthmanT: you don't want this, you want to override self.use_external_resource_naively.

EarthmanT:
Made comments. Resolve them. Squash commits.

PR #156, opened 2016-07-18
CFY-5428 added an instance-subnet connected_to relationship

Jessiebr:


EarthmanT:
Add to this line an or statement to check for this relationships type.

https://github.com/cloudify-cosmo/cloudify-aws-plugin/blob/157e897dabfc5ad8142124e494d1d3808ff381c0/cloudify_aws/ec2/instance.py#L458

EarthmanT:
replacing with a cleaner branch

PR #161, opened 2016-08-13
S3 and ECS poc

geokala:


EarthmanT:
Awesome! Looks like a lot of work. We just merged the restructuring of the plugin. The ec2 and vpc packages now sit inside of the cloudify_aws package. In the root of that folder there is a base.py and a utils.py and a connection.py. Each of those provide functions that the rest of the plugin uses.

Please refactor this branch so that each node_type and relationship is based on either AwsBase (e.g. AwsBaseRelationship, AwsBaseNode). See cloudify_aws/vpc/routetable.py for a pretty thorough example.

In the course of refactoring, you may find that there is an overlap between functionality that you have written and that already exists, e.g. ECSConnectionClient can extend EC2ConnectionClient and reuse _get_aws_config_property. If you see that you can improve base class code, then do. The goal is that we make as much reusable code as possible. This also seriously simplifies writing tests. Use your judgement, if you find improvements that need their own project to implement, create a Jira issue and we'll schedule the improvement. 

There are also some patterns that I want each operation to follow. Look at the code to figure it out. 

Just notice that each module should begin with the operations that are mapped in that module. Those operations should pass a generic `args` argument that will then be passed to the mapped function. The purpose is for a user to override arguments sent to the AWS API with some runtime values. Inside of that function, if relevant, you should look for the appropriate time to override the args sent to the API call. for example, foo.py:
```python
## COPYWRITE BLAH BLAH

imports xyz

@operation
def create(args=None):
    return Taco().create(args)

class Taco(AwsBaseNode):
    def __init__():
        pass

    def create(args):
        create_args = dict(arg1=ctx.node.property['foo']
        create_args = utils.update_args(create_args, args)
        taco = self.execute(self.client.fn, create_args)
```

There are some basic functions to be aware of:

All API communication is accomplished via self.execute(). Getting information about the API object is accomplished by self.get_resource() or self.get_source_resource(). These require the implementation of the get_all_handler. Familiarize yourself with those first. 

Ask me on slack if you have questions.

cloudify_aws/s3/bucket.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def create(self, args=None):
+        self.validate_bucket_name(ctx.node.properties['name'])
+
+        self.execute(
+            self.client.create_bucket,
+            {
+                'Bucket': ctx.node.properties['name'],
+                'ACL': ctx.node.properties['permissions'],
+            }
```
EarthmanT: I prefer that we will not rename abstractions in the properties. if the api is expecting "bucket" and "ACL" then the users should see that in their experience as well. ..also the argument override thing as I mentioned above should be used generally in all operations.

cloudify_aws/s3/bucket.py
```diff
@@ -0,0 +1,178 @@
...
+            raise NonRecoverableError(
+                'For the bucket to be configured as a website, both '
+                'website_index_page and website_error_page must be set.'
+            )
+        else:
+            if '/' in index:
+                raise NonRecoverableError(
+                    'S3 bucket website default page must not contain a /'
+                )
+            self.execute(
```
EarthmanT: Does it make sense to use this as the configure or start operation. I do not like having two api methods in the same operation.

cloudify_aws/s3/object.py
```diff
@@ -0,0 +1,171 @@
...
+def create(args=None, **_):
+    return Object().created(args)
+
+
+@operation
+def delete(args=None, **_):
+    return Object().deleted(args)
+
+
+class Object(AwsBaseNode):
```
EarthmanT: maybe S3Object

cloudify_aws/s3/object.py
```diff
@@ -0,0 +1,171 @@
...
+    def create(self, args=None):
+        bucket_name, bucket_permissions = self._get_bucket_details()
+
+        object_acl = ctx.node.properties.get('permissions', bucket_permissions)
+
+        # Get the contents var
+        is_file = False
+        if ctx.node.properties['load_contents_from_file']:
+            # It's a file, get a file handle for it
+            contents = ctx.download_resource(ctx.node.properties['contents'])
```
EarthmanT: I'm really not crazy about this. A node property should be an abstraction of the resource's static properties. On the other hand, there isn't a good 'Cloudify'-y way of doing things here. The most correct way would be to provide a separate node_template that represents a file object somewhere and to use it, but that's really hard from the user's perspective. Perhaps a good compromise would be a "contents" property and that property will be a new data type with a type field, where you specify "file" or "inline". What do you think?

cloudify_aws/s3/object.py
geokala: That goes against the pattern that we have for all the other entities, and the real problem with 'object' is actually the lower case one. I did consider calling the file s3_object, but if it's enough of a problem then just importing s3 clears up the name issue.

cloudify_aws/s3/bucket.py
geokala: I think in this case the abstraction makes more sense. It'll only be even slightly jarring to someone who is used to the AWS API, and since we're creating an abstraction layer that's not something I'd say we're generally expecting.

In any other cases, having to set the property of 'bucket' on a bucket to set its name feels really peculiar.

I could change permissions -> ACL, but I think for most people, permissions is a bit clearer. Still, let me know if you'd prefer it changed.

cloudify_aws/ecs/service.py
```diff
@@ -0,0 +1,320 @@
...
+        cluster_arn = self.get_cluster_arn(self.get_arn(
+            ctx.instance.relationships,
+            'cloudify.aws.relationships.ecs_service_running_on_cluster',
+        ))
+        task_arn = self.get_arn(
+            ctx.instance.relationships,
+            'cloudify.aws.relationships.ecs_service_runs_task',
+        )
+
+        if None in (cluster_arn, task_arn):
```
geokala: This is changed in the code, but apparently the new way of marking change requests doesn't update when the code is updated so I'm marking it with a comment so that I can see what I have left to deal with.

cloudify_aws/ecs/service.py
```diff
@@ -0,0 +1,320 @@
...
+            ctx.instance.relationships,
+        )
+        if container_instances is None or len(container_instances) == 0:
+            return ctx.operation.retry(
+                'Waiting for cluster to have available instances.',
+            )
+
+        cluster_arn = self.get_cluster_arn(self.get_arn(
+            ctx.instance.relationships,
+            'cloudify.aws.relationships.ecs_service_running_on_cluster',
```
geokala: This is changed in the code, but apparently the new way of marking change requests doesn't update when the code is updated so I'm marking it with a comment so that I can see what I have left to deal with.

cloudify_aws/ecs/service.py
```diff
@@ -0,0 +1,320 @@
...
+            ),
+            desired_attribute='container_names',
+        )
+
+    def create(self, args=None, **_):
+        # Don't try to do anything until the cluster has instances...
+        container_instances = self.get_container_instance_count(
+            ctx.instance.relationships,
+        )
+        if container_instances is None or len(container_instances) == 0:
```
geokala: This is changed in the code, but apparently the new way of marking change requests doesn't update when the code is updated so I'm marking it with a comment so that I can see what I have left to deal with.

cloudify_aws/ecs/cluster.py
```diff
@@ -0,0 +1,237 @@
...
+            client=client,
+            resource_id_key=constants.ECS_CLUSTER['RESOURCE_ID_KEY'],
+        )
+
+        self.not_found_error = constants.ECS_CLUSTER['NOT_FOUND_ERROR']
+
+    def create(self, args=None, **_):
+        cluster_name = ctx.node.properties['name']
+
+        response = self.execute(
```
geokala: This is changed in the code, but apparently the new way of marking change requests doesn't update when the code is updated so I'm marking it with a comment so that I can see what I have left to deal with.

cloudify_aws/s3/bucket.py
geokala: I think of those it makes most sense as a configure operation. I'll switch it to that.

cloudify_aws/ecs/service.py
```diff
@@ -0,0 +1,320 @@
...
+        super(Service, self).__init__(
+            constants.ECS_SERVICE['AWS_RESOURCE_TYPE'],
+            constants.ECS_SERVICE['REQUIRED_PROPERTIES'],
+            client=client,
+            resource_id_key=constants.ECS_SERVICE['RESOURCE_ID_KEY'],
+        )
+
+        self.not_found_error = constants.ECS_SERVICE['NOT_FOUND_ERROR']
+
+    def get_relationship_target(self, relationships, intended_relationship):
```
geokala: I'll take a look at these this evening and migrate them to wherever they seem to fit best.

cloudify_aws/s3/object.py
geokala: Making the contents property into something that has a type sounds sensible. I'll look at that this evening as well.

geokala: Okay, on starting to play with a new datatype, the structure I came up with felt clunky, so I've migrated it to instead have two mutually exclusive properties:
contents
filename

Hopefully that sounds reasonable. Testing at the moment.

geokala:
Okay, I think I've addressed all of your code review comments, other than the ones I queried.

PR #187, opened 2016-09-21
Adding VPC resource create retry support

01000101:
I've been running into race condition issues when creating full network environments in AWS and it's because some resources are being put into a "pending" state after running the create operation and they're not fully available when the next operation attempts to use the resource.  I put in a hook for using the retry operation as well as implemented it in the Subnet create override.


EarthmanT:
Merged to new system tests branch + master

PR #197, opened 2017-02-15
adding secret store support

EarthmanT:


EarthmanT:
Known issue with this PR is that master version of cloudify-plugins-common does not yet include cloudify/plugins/secret_store.py.

EarthmanT:
Closing because we are moving forward with a non-plugin oriented approach.

PR #200, opened 2017-02-16
increase instance tests coverage

Jessiebr:


cloudify_aws/ec2/instance.py
```diff
@@ -133,15 +133,7 @@ def start(self, args=None, start_retry_interval=30,
...
-        if self._get_instance_state() == constants.INSTANCE_STATE_STARTED:
-            if ctx.node.properties['use_password']:
-                password_success = self._retrieve_windows_pass(
-                        instance_id=instance_id,
-                        private_key_path=private_key_path)
-                if not password_success:
-                    return False
-            return True
-
+        self._check_if_instance_started(instance_id, private_key_path)
```
EarthmanT:
```
if self._check_if_instance_started(instance_id, private_key_path):
    return True
```

cloudify_aws/ec2/instance.py
```diff
@@ -155,6 +147,10 @@ def start(self, args=None, start_retry_interval=30,
 
         ctx.logger.debug('Attempted to start instance {0}.'
                          .format(instance_id))
+        self._check_if_instance_started(instance_id, private_key_path)
```
EarthmanT:
```
if self._check_if_instance_started(instance_id, private_key_path):
    return True
```

EarthmanT:
Cherry-picked to AWS-79-1.4.4

PR #214, opened 2017-03-21
AWS-84 Port docs.getcloudify.org docs to Sphinx

funkyHat:


funkyHat:
I believe I've ported all relevant docs across from `aws.md`. The only substantial parts I've not copied are:

```
## cloudify.aws.nodes.KeyPair

**Properties:**

  * `resource_id` Special behavior: when using a new (not external) security group or key pair, this will become the resource name. See [using existing resources section](#using-existing-resources).
```
Is this important to note as "special behaviour" (the detail itself is included in the updated `plugin.yaml`)?


```
The following behaviors are unique:

 * `aws.ec2.instance.start` If `use_external_resource` is true, the `runtime_properties` for `public_ip_address`, etc, are set, and the function exits.
 * `aws.ec2.instance.stop` If `use_external_resource` is true, the `runtime_properties` for `public_ip_address`, etc, are unset, and the function exits.
 * `cloudify.aws.relationships.instance_connected_to_elastic_ip` Here, both the instance's and the `elasticip`'s `use_external_resource` value are relevant. If both are external the function sets the relationship properties. If either are not external the function fails.
 * `cloudify.aws.relationships.instance_connected_to_security_group` Here, both the instance's and the security groups's `use_external_resource` value are relevant. If both are external the function sets the relationship properties. If either are not external the function fails.
 * `cloudify.aws.relationships.route_table_of_source_vpc_connected_to_target_peer_vpc` will run regardless of whether the source or target node is external.
```

I don't feel like this is particularly relevant to users. Shout if you disagree.

docs/types.rst
```diff
@@ -0,0 +1,318 @@
...
+    If you want to create a security group in a VPC,
+    you need to connect it to a VPC using the
+    :cfy:rel:`cloudify.aws.relationships.security_group_contained_in_vpc`
+    relationship.
+
+
+.. cfy:node:: cloudify.aws.nodes.ElasticIP
+
+.. note::
+    the actual IP is available via the ``aws_resource_id`` runtime-property.
```
funkyHat: This seems weird and inconsistent, is it true?

docs/types.rst
```diff
@@ -0,0 +1,318 @@
...
+   It inserts the key's name property in the ``key_name`` parameter in the `run_instances` function.
+
+
+.. cfy:rel:: cloudify.aws.relationships.instance_connected_to_security_group
+
+   The `run_instances` operation looks to see if there are any relationships that define a relationship between the instance and a security group.
+   If so, that security group's ID will be the included in the list of security groups in the ``security_group_ids`` parameter in the `run_instances` function.
+
+
+.. cfy:rel:: cloudify.aws.relationships.instance_connected_to_subnet
```
funkyHat: Why are there 3 of these, and which is the one that should be used?
(I assume it's not cloudify.aws.relationships.instances_contained_in_subnet)

docs/types.rst
```diff
@@ -0,0 +1,318 @@
...
+
+
+.. cfy:node:: cloudify.aws.nodes.ACL
+
+
+.. cfy:node:: cloudify.aws.nodes.DHCPOptions
+
+
+.. cfy:node:: cloudify.aws.nodes.RouteTable
+.. cfy:node:: cloudify.aws.nodes.SecurityGroupRule
```
funkyHat: I found a good number of node_types and relationships were not previously documented.
Please let me know if that was intentional. If so I will need to create some capability for `sphinxify` to mark types as hidden.
(Perhaps what we really need is a way to mark types as deprecated within the DSL?)

circle.yml
```diff
@@ -15,8 +15,8 @@ dependencies:
 
 test:
   override:
-    - tox -e flake8
```
funkyHat: No point splitting up the `tox` call if you're not going to pass any args in or anything

docs/Makefile
```diff
@@ -0,0 +1,225 @@
+# Makefile for Sphinx documentation
```
funkyHat: Boilerplate generated by `sphinx-quickstart` (keeping it here allows developers to simply run `make html` from the `docs/` dir to test building the docs)

docs/changelog.rst
```diff
@@ -0,0 +1,13 @@
...
+
+Changelog
+=========
+
+.. literalinclude:: ../CHANGELOG.txt
+    :language: yaml
+
+
+# Contributors
+# ============
```
funkyHat: Commented because we don't currently have a CONTRIBUTORS file for AWS

plugin.yaml
```diff
@@ -191,27 +206,39 @@ node_types:
...
-          The instance's size.
+          The instance type for the instance.
+          This describes the instance's size.
+          For more information, refer to
+          http://aws.amazon.com/ec2/instance-types/.
         type: string
         required: true
       use_password:
+        description: >
+          Not used by cloudify.aws.nodes.Instance.
```
funkyHat: Is this true? I couldn't find it in the code

docs/types.rst
```diff
@@ -0,0 +1,318 @@
...
+   Otherwise, the instance is in the EC2 Classic VPC.
+
+.. cfy:rel:: cloudify.aws.relationships.connected_to_subnet
+
+   TODO: WAT
+
+
+.. cfy:rel:: cloudify.aws.relationships.instance_contained_in_subnet
+.. warning::
+   Deprecated! Please use
```
funkyHat: I have just now decided that this relationship is deprecated. Please shout if I'm wrong (but if I'm wrong we should probably have a discussion about it)

docs/types.rst
EarthmanT: I'm sure that underlying this diversion of the norm is some very interesting reason, but not interesting enough to warrant the search for it.

docs/types.rst
EarthmanT: Using SecurityGroupRule as an example, I'd say that this is do to negligence in updating the docs, and not to any intention to hide types.

docs/types.rst
EarthmanT: Backwards compatibility. After several iterations we found that it's best to have a generic "connected_to_subnet" relationship that many operations can use internally. Previous versions use "cloudify.aws.relationships.instances_contained_in_subnet" which introduced a bug that instance could only be attached to one subnet, and "cloudify.aws.relationships.instance_connected_to_subnet" which just allows instances to use the relationships.

plugin.yaml
EarthmanT: https://github.com/cloudify-cosmo/cloudify-aws-plugin/blob/master/cloudify_aws/ec2/instance.py#L143

docs/types.rst
EarthmanT: You are right.

plugin.yaml
funkyHat: Ah, sorry, my grep-fu has failed me and I somehow didn't find this

PR #216, opened 2017-04-03
Fixing validation errors

Jessiebr:


cloudify_aws/ec2/instance.py
```diff
@@ -655,7 +655,10 @@ def _get_image(self, image_id):
         return image_object
 
     def get_resource(self):
-        return self._get_instance_from_id(self.resource_id)
```
EarthmanT: We should not modify the get_resource function. This function is used by other functions in the plugin and they rely on it's current behavior.
It is better to try to make the change inside of the creation_validation operation. Maybe it needs to have its own function for checking resources.

PR #234, opened 2017-07-03
fix instance get resource with no id returns something

Jessiebr:


cloudify_aws/ec2/instance.py
```diff
@@ -531,7 +531,8 @@ def _get_instance_from_id(self, instance_id):
 
         instance = self._get_all_instances(list_of_instance_ids=instance_id)
 
-        return instance[0] if instance else instance
+        return instance[0] if instance and instance[0].id == instance_id \
+            else None
```
EarthmanT: Even though it was like this, first "instance[0]" this isn't very good because you can easily get an IndexError.
This is probably the most similar thing that wont result in an index error:
```python
    return None if not instance else instance[0] if instance[0].id == instance_id else None
```

PR #242, opened 2017-09-20
AWS-109: first get parameters from operation inputs, then handle user…

isaac-s:
…data

### DIFF
--- a/cloudify_aws/ec2/instance.py
+++ b/cloudify_aws/ec2/instance.py
@@ -445,8 +445,8 @@ class Instance(AwsBaseNode):
         })
 
         parameters.update(ctx.node.properties['parameters'])
-        parameters = self._handle_userdata(parameters)
         parameters = utils.update_args(parameters, args)
+        parameters = self._handle_userdata(parameters)
         parameters['block_device_map'] = \
             self._create_block_device_mapping(
                 parameters.get('block_device_map', {})


EarthmanT:
The purpose of inputs override is to **override**. That aside I do see the value in being able to provide a user data at runtime and install the agent via user data. What I would prefer is an example of a script that installs the agent via user data so that users can combine the override with the user data instead of breaking the convention of inputs being an override approach, which is very valuable.

isaac-s:
@EarthmanT if operation inputs are indeed supposed to override, then the OpenStack plugin suffers from the same inconsistency. In OpenStack, we handle userdata after incorporating operation arguments.

So:

* In OpenStack's case, we treat the `args` as inputs override for everything *except* userdata.
* In the current AWS plugin, we treat `args` as inputs override for everything *including* userdata.

Both should be consistent. Either one should align with the other.

If we stick with `args` being strictly overrides for everything, then perhaps we should add another operation input called `userdata`, which is what users would use in order to provide userdata during runtime. Then, `args` can be treated as a strict override.

EarthmanT:
@isaac-s 

I'm not suggesting it will be one or the other. Having all other inputs be overrides, aside from user data is inconsistent. There also is no need for a special "user data" input, which is even more messy.

I have something else in mind, that is hopefully the best of both worlds:

- If the agent config specifies user data agent installation, then the plugin should know that this is a priority.
- Ideally the function should be able to combine user agent install script with the inputs override user data. This can be simple concatenation. The question is whether the agent installation should be prepended or appended to the user data override.
- We should also take into account the superiority of "cloud config" cloud init format, and also have a similar handling for that scenario.

Thoughts?

EarthmanT:
This is in 1.5.1 branch.

PR #260, opened 2019-01-04
Add region check before create client for ec2/rds

0lvin:


cloudify_aws/common/utils.py
```diff
@@ -457,11 +456,29 @@ def get_tags_list(node_prop, runtime_prop, input_prop):
     return tags_list
 
 
+def check_region_name(region):
+    region_matcher = re.compile(constants.REGION_REGEX)
+    if not region_matcher.match(region):
+        raise NonRecoverableError(
+            'Looks as wrong region name:{}'
```
EarthmanT: 1. Let's change the message to "The `region_name` provided in `client_config` does not match the expected region regular expression."
2. Use an index in the format string method, i.e. "{0}".

cloudify_aws/common/utils.py
```diff
@@ -457,11 +456,29 @@ def get_tags_list(node_prop, runtime_prop, input_prop):
...
+            .format(repr(region)))
+
+
+def check_availability_zone(zone):
+    zone_matcher = re.compile(constants.AVAILABILITY_ZONE_REGEX)
+    if not zone_matcher.match(zone):
+        raise NonRecoverableError(
+            'Looks as wrong availability zone: {}'
+            .format(repr(zone)))
```
EarthmanT: 1. Let's change the message to "The `AvailabilityZone` provided in `resource_config` does not match the expected availability zone regular expression."
2. Use an index in the format string method, i.e. "{0}".

cloudify_aws/common/utils.py
```diff
@@ -457,11 +456,29 @@ def get_tags_list(node_prop, runtime_prop, input_prop):
...
+
 def clean_params(p):
     if not isinstance(p, dict) or not p:
         return {}
     for _k, _v in p.items():
         if not _v:
             del p[_k]
         del _k, _v
+    if 'AvailabilityZone' in p:
+        check_availability_zone(p['AvailabilityZone'])
```
EarthmanT: If we are already looping through `p` above, why not check the condition in the loop, instead of having it iterate p again?

cloudify_aws/ec2/__init__.py
```diff
@@ -29,6 +30,10 @@ class EC2Base(AWSResourceBase):
         AWS ELB base interface
     """
     def __init__(self, ctx_node, resource_id=None, client=None, logger=None):
+        if not client:
+            aws_config = ctx_node.properties.get(AWS_CONFIG_PROPERTY, dict())
+            check_region_name(aws_config['region_name'])
+
```
EarthmanT: Why do we only run this if the `client` is `None`?

0lvin: If we already have client - i suppose that aws already checked credentials.

PR #274, opened 2019-07-04
Resumable actions support

0lvin:


cloudify_aws/autoscaling/resources/launch_configuration.py
```diff
@@ -62,9 +62,6 @@ def properties(self):
     @property
     def status(self):
         """Gets the status of an external resource"""
-        props = self.properties
-        if not props:
-            return None
         return None
```
EarthmanT: Why? (Applies to other files.)

0lvin: Code return None for any state of properties. I have not found use cases for get properties without any usage.

PR #278, opened 2019-07-30
Cy 1561 add ebs relationship for ec2

mabuaisha:


cloudify_aws/ec2/resources/ebs.py
```diff
@@ -204,58 +248,95 @@ def delete(ctx, iface, resource_config, **_):
...
+    utils.update_resource_id(ctx.instance, esp_id)
+    iface.update_resource_id(esp_id)
 
 
 @decorators.aws_resource(EC2VolumeAttachment, RESOURCE_TYPE_VOLUME_ATTACHMENT,
                          ignore_properties=True)
 @decorators.wait_for_status(status_good=[DETACHED, AVAILABLE],
                             status_pending=[DETACHING, INUSE])
-def detach(ctx, iface, resource_config, **_):
+def delete_attachment(ctx, iface, **_):
```
EarthmanT: We don't need to keep the old name, but we should have a function with the old name that calls the new function in case someone is using it. OK?

setup.py
```diff
@@ -19,7 +19,7 @@
 
 setup(
     name='cloudify-aws-plugin',
-    version='2.2.1',
+    version='2.2.2',
     author='Cloudify Platform Ltd.',
```
EarthmanT: should 2.3.0.

CHANGELOG.txt
```diff
@@ -65,3 +65,5 @@
     - Add resumable support to operations
 2.2.1:
     - Fix bug when delete EC2 instance
+2.2.2:
+    - Add new EBS relationship to attach ebs volume to EC2 instance
```
EarthmanT: Add message about changing in handling on delete (issue 128 and issue 129)

cloudify_aws/ec2/resources/ebs.py
```diff
@@ -82,6 +84,23 @@ def status(self):
         return self.properties[VOLUME_STATE]\
             if self.properties and self.properties.get(VOLUME_STATE) else None
 
+    def attach(self, params):
+
+        return self.make_client_call('attach_volume', params)
+
+    def detach(self, params=None):
```
EarthmanT: maybe we should add `params = params or {}` to avoid a TypeError on `**params`?

cloudify_aws/ec2/resources/ebs.py
mabuaisha: Make sense

setup.py
mabuaisha: Done

cloudify_aws/ec2/resources/ebs.py
mabuaisha: Done

CHANGELOG.txt
mabuaisha: Can you elaborate  more on this ?

PR #285, opened 2020-03-09
CY-2266-Support-EKS-With-Boto3-Upgrade

ahmadiesa-abu:
In order to support [ create node_group -needed to create nodes in the EKS cluster- ] , we had to upgrade boto since the old one that we were using only support create EKS cluster only , without the node-group

old : https://boto3.amazonaws.com/v1/documentation/api/1.9.75/reference/services/eks.html
vs
new : https://boto3.amazonaws.com/v1/documentation/api/1.12.13/reference/services/eks.html

### DIFF
--- a/cloudify_aws/eks/resources/cluster.py
+++ b/cloudify_aws/eks/resources/cluster.py
@@ -1,0 +1,227 @@
+# Copyright (c) 2018 Cloudify Platform Ltd. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


examples/eks-feature-demo/blueprint.yaml
```diff
@@ -0,0 +1,580 @@
+tosca_definitions_version: cloudify_dsl_1_3
```
EarthmanT: I'd like to add this example to cloudify-community/blueprint-examples/kubernetes

cloudify_aws/eks/resources/cluster.py
```diff
@@ -0,0 +1,227 @@
...
+    if response and response.get(CLUSTER):
+        resource_arn = response.get(CLUSTER).get(CLUSTER_ARN)
+        utils.update_resource_arn(ctx.instance, resource_arn)
+    # wait for cluster to be active
+    ctx.logger.info("Waiting for Cluster to become Active")
+    iface.wait_for_cluster(params, 'cluster_active')
+    if store_kube_config_in_runtime:
+        client_config = ctx.node.properties['client_config']
+        ctx.instance.runtime_properties['kubeconf'] = \
+            iface.get_kubeconf(client_config, params)
```
EarthmanT: Is there anychance that this could contain some object that is not serializable in JSON?

PR #298, opened 2020-07-05
CY-2801-Availability-Zone-Retry

ahmadiesa-abu:
added a simple check before creating the resource that takes Availability zone as input

### DIFF
--- a/cloudify_aws/ec2/resources/ebs.py
+++ b/cloudify_aws/ec2/resources/ebs.py
@@ -235,12 +235,14 @@ def create(ctx, iface, resource_config, **_):
 
     # Actually create ebs resource
     region_name = ctx.node.properties['client_config']['region_name']
-    use_available_zones = params.pop('UseAvailableZones', False)
+    use_available_zones = ctx.node.properties.get('use_available_zones', False)
     try:
         create_response = iface.create(params)
     except CapacityNotAvailableError:
-        ctx.logger.info("The Availability Zone chosen is not available")
         if use_available_zones:
+            ctx.logger.warn(
+                "The Availability Zone chosen {0} "
+                "is not available".format(params['AvailabilityZone']))
             valid_zone = \
                 iface.get_available_zone({
                     'Filters': [
@@ -248,6 +250,8 @@ def create(ctx, iface, resource_config, **_):
                     ]
                 })
             if valid_zone:
+                ctx.logger.info(
+                    "using {0} Availability Zone instead".format(valid_zone))
                 params['AvailabilityZone'] = valid_zone
                 create_response = iface.create(params)
             else:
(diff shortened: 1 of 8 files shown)


cloudify_aws/ec2/resources/ebs.py
```diff
@@ -232,8 +233,30 @@ def create(ctx, iface, resource_config, **_):
     params = utils.clean_params(
         dict() if not resource_config else resource_config.copy())
 
-    # Create ebs resource
-    create_response = iface.create(params)
+    # Actually create ebs resource
+    region_name = ctx.node.properties['client_config']['region_name']
+    use_available_zones = params.pop('UseAvailableZones', False)
```
EarthmanT, line 238: Should be `ctx.node.properties['use_available_zones']`, or is that a real parameter for create?

cloudify_aws/ec2/resources/ebs.py
```diff
@@ -232,8 +233,30 @@ def create(ctx, iface, resource_config, **_):
...
+    try:
+        create_response = iface.create(params)
+    except CapacityNotAvailableError:
+        ctx.logger.info("The Availability Zone chosen is not available")
```
EarthmanT, line 242: Add which zone failed in the message.

EarthmanT: Maybe it should be WARN level. `ctx.logger.warn`.

cloudify_aws/ec2/resources/subnet.py
```diff
@@ -143,7 +145,30 @@ def create(ctx, iface, resource_config, **_):
...
+                params['AvailabilityZone'] = valid_zone
+                create_response = iface.create(params)[SUBNET]
+            else:
+                raise NonRecoverableError(
+                    "no available Availability Zones "
+                    "in region {0}".format(region_name))
+        else:
+            raise NonRecoverableError(
+                "The Availability Zone chosen is not available")
+
```
EarthmanT, line 171: Same comments apply here.

plugin.yaml
```diff
@@ -2,9 +2,9 @@ plugins:
 
   aws:
     executor: central_deployment_agent
-    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/2.3.5.zip
+    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/2.3.6.zip
     package_name: cloudify-aws-plugin
-    package_version: '2.3.5'
+    package_version: '2.3.6'
```
EarthmanT, line 7: Let's make it 2.4. It's a feature not a bug.

plugin.yaml
```diff
@@ -651,6 +651,10 @@ data_types:
         type: string
         description: The Availability Zone for the subnet.
         required: false
+      UseAvailableZones:
+        type: boolean
+        description: A boolean to choose another available zone if the one provided is not available.
+        required: false
```
EarthmanT, line 657: `resource_config` is reserved for API keywords, this should be a property. And properties are all lower case separated by `_`.

cloudify_aws/ec2/resources/ebs.py
ahmadiesa-abu: @EarthmanT  we added that to the properties , and if passed to the API it will fail , so I just pop it to remove it from being passed to the API

PR #308, opened 2020-10-13
handle nonstring tags

EarthmanT:
__NODOCS__


cloudify_aws/common/utils.py
```diff
@@ -468,9 +468,10 @@ def get_tags_list(node_prop, runtime_prop, input_prop):
     if isinstance(node_prop, list):
         tags_list = node_prop
     if isinstance(runtime_prop, list):
-        tags_list = list(set(tags_list + runtime_prop))
+        tags_list.extend(runtime_prop)
```
AdarShaked, line 471: What happens if the same tag appears in `tags_list` and `runtime_prop` ?
its ok to have same tag twice in the list?

AdarShaked (approved):

PR #309, opened 2020-11-12
Fix payload in request and reponse for Lambda call

kfrukacz-gs:
Payload payload bytes are read, but this cannot be dumped into runtime property.
This code adds parsing of the response payload (and additionally data inside which also is json)

### DIFF
--- a/cloudify_aws/lambda_serverless/resources/function.py
+++ b/cloudify_aws/lambda_serverless/resources/function.py
@@ -95,12 +95,11 @@ class LambdaFunction(LambdaBase):
                 encoded_payload:
             if encoded_payload is not None:
                 invoke_params['Payload'] = encoded_payload
-        self.logger.debug('Invoking %s with parameters: %s'
-                          % (self.type_name, invoke_params))
-        res = self.client.invoke(**invoke_params)
-        payload = self._decode_payload(res.get('Payload'))
-        if payload:
-            res['Payload'] = payload
+            self.logger.debug('Invoking %s with parameters: %s'
+                              % (self.type_name, invoke_params))
+            res = self.client.invoke(**invoke_params)
+        if res and res.get('Payload'):
+            res['Payload'] = self._decode_payload(res['Payload'])
         self.logger.debug('Response: %s' % res)
         return res
 
@@ -109,23 +108,22 @@ class LambdaFunction(LambdaBase):
         if isinstance(payload, str):
             with file(payload, 'r') as payload_file:
                 yield payload_file
-        if isinstance(payload, dict):
+        elif isinstance(payload, dict):
             yield json.dumps(payload).encode(self.resource_encoding)
-        yield payload
+        else:
+            yield payload
 
     def _decode_payload(self, payload_stream):
-        payload = None
-        if payload_stream:
-            try:
-                payload = json.loads(
-                    payload_stream.read().decode(self.resource_encoding))
-                if payload.get('body'):
-                    try:
-                        payload['body'] = json.loads(payload['body'])
-                    except ValueError as e:
-                        self.logger.warn('Cannot decode body: %s' % e)
-            except (ValueError, UnicodeDecodeError) as e:
-                self.logger.warn('Cannot decode Payload: %s' % e)
+        payload = payload_stream.read().decode(self.resource_encoding)
+        try:
+            payload = json.loads(payload)
+            if payload.get('body'):
+                try:
+                    payload['body'] = json.loads(payload['body'])
+                except ValueError as e:
+                    self.logger.warn('Cannot decode body: %s' % e)
+        except (ValueError, UnicodeDecodeError) as e:
+            self.logger.warn('Cannot decode Payload: %s' % e)
         return payload
 
 
--- a/cloudify_aws/lambda_serverless/resources/function.py
+++ b/cloudify_aws/lambda_serverless/resources/function.py
@@ -81,13 +81,15 @@ class LambdaFunction(LambdaBase):
         '''
             Invokes an AWS Lambda Function.
         '''
-        params = params or dict()
-        params.update(dict(FunctionName=self.resource_id))
-        if params.get('Payload'):
-            params['Payload'] =  _encode_payload(params['Payload'])
+        invoke_params = dict()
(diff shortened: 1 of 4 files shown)


cloudify_aws/lambda_serverless/resources/function.py
```diff
@@ -85,11 +87,16 @@ def invoke(self, params):
...
         res = self.client.invoke(**params)
         if res and res.get('Payload'):
-            res['Payload'] = res['Payload'].read()
+            try:
+                res['Payload'] = res['Payload'].read()
+                res['Payload'] = json.loads(res['Payload'].decode('utf-8'))
+                if res['Payload'].get('body'):
+                    res['Payload']['body'] = json.loads(res['Payload']['body'])
+            except json.JSONDecodeError:
+              pass
```
mabuaisha, line 96: @kfrukacz-gs You handled the response, but what about request ? does the json string passed correctly when you use paylaod from runtime properties ? and we need also to update the file-like object for payload

mabuaisha (commented):
left a comment

cloudify_aws/lambda_serverless/resources/function.py
kfrukacz-gs: @mabuaisha  working on it

cloudify_aws/lambda_serverless/resources/function.py
```diff
@@ -83,6 +83,8 @@ def invoke(self, params):
         '''
         params = params or dict()
         params.update(dict(FunctionName=self.resource_id))
+        if params.get('Payload'):
+            params['Payload'] =  _encode_payload(params['Payload'])
```
mabuaisha, line 87: @kfrukacz-gs sounds good, to make this also work with file-like object we should also add support for that

mabuaisha:
@EarthmanT: @kfrukacz-gs found a bug on the aws plugin related to invoke labmda function that needs payload passed as input for both string/dict/file-like object and he already fixed one of the issue and there is another one currently he is updating

cloudify_aws/lambda_serverless/resources/function.py
```diff
@@ -79,16 +81,32 @@ def invoke(self, params):
         '''
             Invokes an AWS Lambda Function.
         '''
-        params = params or dict()
-        params.update(dict(FunctionName=self.resource_id))
+        invoke_params = dict()
+        invoke_params.update(params)
+        invoke_params.update(dict(FunctionName=self.resource_id))
+        if invoke_params.get('Payload'):
```
EarthmanT, line 87: Prefer
```
payload = invoke_params.get('Payload')
if payload:
    ----
```

plugin.yaml
```diff
@@ -2,7 +2,6 @@ plugins:
 
   aws:
     executor: central_deployment_agent
-    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/2.4.4.zip
```
mabuaisha, line 5: @kfrukacz-gs why did you drpp the source ?
and more thing we need to increment the 2.4.4 to 2.4.5 and also check the setup.py as well

cloudify_aws/lambda_serverless/resources/function.py
```diff
@@ -79,16 +88,44 @@ def invoke(self, params):
...
 
+    @contextmanager
+    def _encode_payload(self, payload):
+        if isinstance(payload, str):
+            with file(payload, 'r') as payload_file:
+                yield payload_file
+        elif isinstance(payload, dict):
+            yield json.dumps(payload).encode(self.resource_encoding)
+        else:
+            yield payload
```
mabuaisha, line 114: @kfrukacz-gs whats the point of using the context manager here ? I mean it can be normal method

cloudify_aws/lambda_serverless/resources/function.py
```diff
@@ -79,16 +88,44 @@ def invoke(self, params):
...
+        payload = payload_stream.read().decode(self.resource_encoding)
+        try:
+            payload = json.loads(payload)
+            if payload.get('body'):
+                try:
+                    payload['body'] = json.loads(payload['body'])
+                except ValueError as e:
+                    self.logger.warn('Cannot decode body: %s' % e)
+        except (ValueError, UnicodeDecodeError) as e:
+            self.logger.warn('Cannot decode Payload: %s' % e)
```
mabuaisha, line 126: Do we really care here about just rasing warning  ?

mabuaisha (commented):
Added some comments

cloudify_aws/lambda_serverless/resources/function.py
kfrukacz-gs: it is used for handling the file open for the boto3 and closing when done

cloudify_aws/lambda_serverless/resources/function.py
kfrukacz-gs: not really, I will remove the warn

plugin.yaml
kfrukacz-gs: ok

cloudify_aws/lambda_serverless/resources/function.py
mabuaisha: I mean you do not close antyhing or do I miss anything here ? with file just close the file but what I meant the context manager does not add anything here, the method can be normal without decorator and still got the same result

kfrukacz-gs: file is opened in line 109 and the stream is yielded, since it is "with" it will be closed after done

cloudify_aws/lambda_serverless/resources/function.py
kfrukacz-gs: done

plugin.yaml
kfrukacz-gs: done

cloudify_aws/lambda_serverless/resources/function.py
kfrukacz-gs: code refactored

kfrukacz-gs:
@mabuaisha @EarthmanT  will this code be merged to the repo?

PR #310, opened 2020-11-23
Rd 317 add pipeline support

AdarShaked:


cloudify_aws/codepipeline/resources/pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def properties(self):
+        """Gets the properties of an external resource"""
+        resources = None
+        try:
+            resources = self.client.get_pipeline_state(
+                name=self.resource_id)
+        except ClientError:
+            pass
+        return resources
```
EarthmanT, line 49: Is this possibly a list of more than one resource? Because below in `status`, we are treating it a dict.

cloudify_aws/codepipeline/resources/pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @property
+    def status(self):
+        """Gets the status of an external resource"""
+        props = self.properties
+        if props:
+            if props.get('created'):
+                return CREATED_STATUS
+        return None
```
EarthmanT, line 58: `None` is redundant. Is it needed?

cloudify_aws/codepipeline/resources/pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def create(self, params):
+        """
+            Create a new Pipeline .
+        """
+        return self.make_client_call('create_pipeline', params)
```
EarthmanT, line 64: Do we want to log params in debug?

cloudify_aws/codepipeline/resources/pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def delete(self, params=None):
+        """
+            Deletes an existing Pipeline.
+        """
+        self.logger.debug('Deleting {resource_type} with parameters:'
+                          ' {params}'.format(
+                                            resource_type=self.type_name,
+                                            params=params))
+        self.client.delete_pipeline(**params)
```
EarthmanT, line 74: Do we want to log result?

plugin.yaml
```diff
@@ -1074,6 +1074,12 @@ data_types:
         description: http://boto3.readthedocs.io/en/latest/reference/services/eks.html#EKS.Client.create_nodegroup
```
EarthmanT, line 1074: update version (and changelog)

EarthmanT (changes requested):

cloudify_aws/codepipeline/resources/pipeline.py
AdarShaked: Helped me to debug and understand what params being passed.

cloudify_aws/codepipeline/resources/pipeline.py
AdarShaked: No

cloudify_aws/codepipeline/resources/pipeline.py
AdarShaked: I don't think its needed in the delete. If there is an error the user will see it.

cloudify_aws/codepipeline/resources/pipeline.py
AdarShaked: No, my fault.

EarthmanT (approved):

PR #313, opened 2020-11-29
RD-650: set LocationConstraint to be optional

isaac-s:


EarthmanT (dismissed):
Please add `__NODOCS__` to commit message and update CHANGELOG.txt and plugin.yaml version.

isaac-s:
@EarthmanT unfortunately this doesn't work as expected. `CreateBucketConfiguration` is being included with the request, as an empty dict, causing a failure. This helps, but I really don't wanna do it:

```
    if 'CreateBucketConfiguration' in resource_config and \
        not resource_config['CreateBucketConfiguration']:
        resource_config.pop('CreateBucketConfiguration')
```

EarthmanT:
> @EarthmanT unfortunately this doesn't work as expected. `CreateBucketConfiguration` is being included with the request, as an empty dict, causing a failure. This helps, but I really don't wanna do it:
> 
> ```
>     if 'CreateBucketConfiguration' in resource_config and \
>         not resource_config['CreateBucketConfiguration']:
>         resource_config.pop('CreateBucketConfiguration')
> ```

If it was me, I would use `del` other than that it seems like the best solution.

EarthmanT (approved):

EarthmanT (approved):

PR #320, opened 2020-12-16
RD-815: Added relationship for main route table

isaac-s:
__NODOCS__

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -404,9 +404,6 @@ jobs:
       TEST_NAME: virtual-machine
     steps:
       - checkout
-      - run:
-          name: wait for starter
-          command: docker exec -it cfy_manager cfy_manager wait-for-starter
       - run_integration_tests_505
 
   eks_integration_tests_505:
@@ -417,9 +414,6 @@ jobs:
       TEST_NAME: eks
     steps:
       - checkout
-      - run:
-          name: wait for starter
-          command: docker exec -it cfy_manager cfy_manager wait-for-starter
       - run_integration_tests_505
 
   hw_integration_tests_505:
@@ -430,9 +424,6 @@ jobs:
       TEST_NAME: hello-world
     steps:
       - checkout
-      - run:
-          name: wait for starter
-          command: docker exec -it cfy_manager cfy_manager wait-for-starter
       - run_integration_tests_505
 
   cf_integration_tests_510:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -404,6 +404,9 @@ jobs:
       TEST_NAME: virtual-machine
     steps:
       - checkout
+      - run:
+          name: wait for starter
+          command: docker exec -it cfy_manager cfy_manager wait-for-starter
       - run_integration_tests_505
 
   eks_integration_tests_505:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -424,6 +430,9 @@ jobs:
       TEST_NAME: hello-world
     steps:
       - checkout
+      - run:
+          name: wait for starter
+          command: docker exec -it cfy_manager cfy_manager wait-for-starter
```
mabuaisha, line 435: The wait-for-starter is available only on 5.1.0 not on 505 so you do not need to have it here

.circleci/config.yml
```diff
@@ -414,6 +417,9 @@ jobs:
       TEST_NAME: eks
     steps:
       - checkout
+      - run:
+          name: wait for starter
+          command: docker exec -it cfy_manager cfy_manager wait-for-starter
```
mabuaisha, line 422: This is only available on 5.1.0 `wait-for-starter` command

.circleci/config.yml
```diff
@@ -404,6 +404,9 @@ jobs:
       TEST_NAME: virtual-machine
     steps:
       - checkout
+      - run:
+          name: wait for starter
+          command: docker exec -it cfy_manager cfy_manager wait-for-starter
```
mabuaisha, line 409: This is only available on 5.1.0 wait-for-starter command

EarthmanT (dismissed):

cloudify_aws/ec2/resources/routetable.py
```diff
@@ -185,3 +188,31 @@ def detach(ctx, iface, resource_config, **_):
...
+        main_route_table_id = None
+        for route_table in route_tables:
+            for association in route_table.get('Associations', []):
+                if association.get('Main'):
+                    main_route_table_id = route_table['RouteTableId']
+        if not main_route_table_id:
+            raise NonRecoverableError(
+                "VPC '%s' does not have a main routing table" % vpc_id)
+        ctx.source.instance.runtime_properties[EXTERNAL_RESOURCE_ID] = \
+            main_route_table_id
```
mabuaisha, line 218: @isaac-s with this change you are overriding the `EXTERNAL_RESOURCE_ID` for the current node template for the routing table node. Basically you are using the node routing able that is contained in a VPC to check what are the routing tables associated with that VPC so that you can get them all and get the one marked as main so that you can get its id. do not you think this is a little bit dangerous if anyone tries to reference the original routing table id then it will be something else since you override it here ?

Cannot you do that on the VPC level by exposing the list of routing table and find out the main one ? unless in your case the vpc is an external then it will be an issue ?

mabuaisha (changes requested):
add some comments

cloudify_aws/ec2/resources/routetable.py
isaac-s: Well, you're right. I can add a check there, and skip this entire thing if `EXTERNAL_RESOURCE_ID` is provided.

Regarding the VPC level: exactly. My original solution was to do this at the VPC level, but that won't work for existing VPC's because there's no callback I can use to populate the `main_route_table_id` runtime property. We spoke about it over the phone...

isaac-s: @mabuaisha I added more strict checks.

mabuaisha (dismissed):

PR #322, opened 2020-12-18
RD-815 v2

isaac-s:


cloudify_aws/common/__init__.py
```diff
@@ -116,3 +116,7 @@ def verify_resource_exists(self):
         if not getattr(self, 'properties', {}):
             return False
         return True
+
```
EarthmanT, line 119: Can we add more to the docstring so that future devs don't need to ask what this is for? Also in the decorator.

EarthmanT (changes requested):

EarthmanT (approved):

EarthmanT:
Save yourself now and add __NODOCS__ to the commit.

EarthmanT (approved):

PR #325, opened 2020-12-22
Added class_decl so we can use populate_resource

isaac-s:


EarthmanT:
update version in plugin yaml/changelog

EarthmanT (approved):

PR #327, opened 2021-01-17
2.5.7 build

EarthmanT:


EarthmanT:
@kfrukacz-gs Py3 tests are failing. Are you sure that the expected response is correct here?
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.invoke
https://botocore.amazonaws.com/v1/documentation/api/latest/reference/response.html

Also please check the change I made `file`-> `open`.

PR #340, opened 2021-03-14
fix device index when using relationships

EarthmanT:


cloudify_aws/ec2/resources/instances.py
```diff
@@ -220,7 +220,7 @@ def create(ctx, iface, resource_config, **_):
                 'device_index')
             rel_nic = {
                 NIC_ID: rel_nic_id,
-                DEVICE_INDEX: rel_device_index
+                DEVICE_INDEX: rel_device_index or counter
```
AdarShaked, line 223: There is a possibility that for nic A target.target.instance.runtime_properties.get(
                'device_index') =2 and for nic B rel_device_index then rel_device_index will be 2 and we end up with two devices with same index?its a problem?

AdarShaked (approved):

PR #371, opened 2021-08-15
add cloudify_tag property to plugin.yaml

EarthmanT:
add default tagging methods to decorators.py

### DIFF
--- a/cloudify_aws/common/decorators.py
+++ b/cloudify_aws/common/decorators.py
@@ -475,17 +475,15 @@ def tag_resources(fn):
         resource_id = utils.get_resource_id(
             node=ctx.node,
             instance=ctx.instance)
-        try:
-            if ctx.node.properties['cloudify_tagging'] == True:
-                tags = get_default_tag(ctx, iface)
-                ctx.logger.info("added deafult cloudify_tagging")
-        except:
-            tags = {}
+        if ctx.node.properties.get('cloudify_tagging', False):
+            add_default_tag(ctx, iface)
+            ctx.logger.info("added default cloudify_tagging")
+        else:
             ctx.logger.info("cloudify_tagging missing in node")
-        tags.update(utils.get_tags_list(
+        tags = utils.get_tags_list(
             ctx.node.properties.get('Tags'),
             ctx.instance.runtime_properties.get('Tags'),
-            kwargs.get('Tags')))    
+            kwargs.get('Tags'))    
         if iface and tags and resource_id:
             iface.tag({
                 'Tags': tags,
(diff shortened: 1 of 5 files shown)


plugin.yaml
```diff
@@ -2,9 +2,9 @@ plugins:
 
   aws:
     executor: central_deployment_agent
-    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/2.12.2.zip
+    source: https://github.com/cloudify-cosmo/cloudify-aws-plugin/archive/2.12.3.zip
     package_name: cloudify-aws-plugin
-    package_version: '2.12.2'
+    package_version: '2.12.4'
```
EarthmanT, line 8: version in line 5 and 7 should be the same.

plugin.yaml
```diff
@@ -2405,6 +2405,9 @@ node_types:
       <<: *client_config
       <<: *resource_id
       <<: *tags_property
+      cloudify_tag:
+        description: an automatic tag to identify the ec2 instance
+        type: boolean
```
EarthmanT, line 2410: In the code, you put `cloudify_tagging`. Which is it, `cloudify_tagging` or `cloudify_tags`? We should also add this to `cloudify.nodes.aws.ec2.BaseType` and other ones as well.

cloudify_aws/common/decorators.py
```diff
@@ -479,6 +479,8 @@ def wrapper(**kwargs):
             ctx.node.properties.get('Tags'),
             ctx.instance.runtime_properties.get('Tags'),
             kwargs.get('Tags'))
+        if ctx.node.properties['cloudify_tagging'] == True:
```
EarthmanT:
```suggestion
        if ctx.node.properties.get('cloudify_tagging', False):
```

When adding a property, I will usually use `.get`, because if someone is using  code that does not have this property, like from an old blueprint, then it will not fail from backward compat. Also, in Python, we don't generally need to do: `== False` or `== None` or `==True`. A boolean evaluation `if` is enough. That's the case unless you know that there is a chance that someone is mixing types, like if we change the type, or if the API doesn't send a boolean, but a string of a boolean value, i.e. `false` instead of `False`.

cloudify_aws/common/decorators.py
```diff
@@ -504,3 +506,7 @@ def wrapper(**kwargs):
                 'Resources': [resource_id]})
         return fn(**kwargs)
     return wrapper
+
+def get_default_tag(ctx, iface):
+    iface.tag({'Key': 'CreatedBy', 'Value': '{}-{}-{}'.format(ctx.tenant_name, \
+                                                              ctx.deployment.id, ctx.instance.id)})
```
EarthmanT: We said that we shoul have one tag call.

EarthmanT:
You can look at the tests what's failing. BTW, you also need to do a PR to docs.getcloudify.co for the new feature :)

PR #377, opened 2021-08-31
handling multiple resources per instance

EarthmanT:


cloudify_aws/common/decorators.py
```diff
@@ -495,17 +537,21 @@ def untag_resources(fn):
     def wrapper(**kwargs):
         ctx = kwargs.get('ctx')
         iface = kwargs.get('iface')
-        resource_id = utils.get_resource_id(
-            node=ctx.node,
-            instance=ctx.instance)
+        if MULTI_ID in ctx.instance.runtime_properties:
```
ahmadiesa-abu, line 540: @EarthmanT  could it have the value [] ? in the properties ? , I saw above it is handled with len as well  , or because of in case of untag we are certain about the values since we are populating it on tag

ahmadiesa-abu (commented):
one minor comment , more of a thought ?

ahmadiesa-abu (approved):

PR #403, opened 2021-11-10
Rd 3358 create ami from ec2

EarthmanT:


cloudify_aws/ec2/resources/image.py
```diff
@@ -77,28 +83,87 @@ def create(self, params):
...
+    ctx.instance.runtime_properties['create_response'] = \
+        utils.JsonCleanuper(create_response).to_dict()
+    iface.update_resource_id(create_response['ImageId'])
+    utils.update_resource_id(ctx.instance, create_response['ImageId'])
+    ctx.instance.runtime_properties['AMI_ImageId'] = create_response['ImageId']
+
+
+@decorators.aws_resource(EC2Image, resource_type=RESOURCE_TYPE)
+def delete(ctx, iface, resource_config, **_):
+    """delete/deregister an AWS EC2 Image"""
```
EarthmanT, line 165: Let's use the wait for status decorator. Describe Images returns State value:`
```
            'State': 'pending'|'available'|'invalid'|'deregistered'|'transient'|'failed'|'error',
```
We should expect `deregistered`, however, the user from the blueprint operation inputs should be able to toggle on off this behavior. like `desired_status` property or something. If the user overrides the value with just `''`, then we will not wait.

cloudify_aws/ec2/resources/image.py
```diff
@@ -77,28 +83,87 @@ def create(self, params):
...
+    if not ctx.instance.runtime_properties.get('use_external_resource'):
+        ctx.instance.runtime_properties['resource_config'] = resource_config
+        iface.prepare_describe_image_filter(resource_config.copy())
+        ami = iface.properties
+        utils.update_resource_id(ctx.instance, ami.get(IMAGE_ID))
+
+
+@decorators.aws_resource(EC2Image, resource_type=RESOURCE_TYPE)
+def create(ctx, iface, resource_config, **_):
+    """Create an AWS EC2 Image"""
```
EarthmanT, line 142: Let's use the `wait_for_status` decorator. Describe Images returns image status:
```
            'State': 'pending'|'available'|'invalid'|'deregistered'|'transient'|'failed'|'error',
```
We should expect `available`, however, the user from the blueprint operation inputs should be able to toggle on off this behavior. like `desired_status` property or something. If the user overrides the value with just `''`, then we will not wait.

cloudify_aws/ec2/resources/image.py
```diff
@@ -77,28 +83,87 @@ def create(self, params):
...
+    if ctx.instance.runtime_properties.get('use_external_resource'):
+        # if use_external_resource there we are using an existing image
+        return
+
+    params = utils.clean_params(
+        dict() if not resource_config else resource_config.copy())
+    if 'InstanceId' not in params:
+        params['InstanceId'] = utils.find_resource_id_by_type(
+            ctx.instance, 'cloudify.nodes.aws.ec2.Instances')
+    params = clean_params(params)
```
EarthmanT, line 152: We should only need to clean params once. Maybe we should do that only here and skip it in line 147.

cloudify_aws/ec2/resources/image.py
```diff
@@ -77,28 +83,87 @@ def create(self, params):
...
+
+    # Actually create the resource
+    create_response = iface.create(params)
+    ctx.instance.runtime_properties['create_response'] = \
+        utils.JsonCleanuper(create_response).to_dict()
+    iface.update_resource_id(create_response['ImageId'])
+    utils.update_resource_id(ctx.instance, create_response['ImageId'])
+    ctx.instance.runtime_properties['AMI_ImageId'] = create_response['ImageId']
```
EarthmanT, line 160: Is `aws_resource_id` in  `     utils.update_resource_id(ctx.instance, create_response['ImageId'])` not sufficient?

cloudify_aws/ec2/resources/image.py
```diff
@@ -77,28 +83,87 @@ def create(self, params):
...
-    ctx.instance.runtime_properties['resource_config'] = resource_config
-    iface = \
-        prepare_describe_image_filter(
-            resource_config.copy(),
-            iface)
-    ami = iface.properties
-    utils.update_resource_id(ctx.instance, ami.get(IMAGE_ID))
+    if not iface.describe_image_filters:
+        return
+    if not ctx.instance.runtime_properties.get('use_external_resource'):
```
EarthmanT, line 133: Shouldn't this be `if ctx.instance.runtime_properties.get('use_external_resource')`?

cloudify_aws/ec2/resources/image.py
```diff
@@ -77,28 +83,87 @@ def create(self, params):
...
+                new_params[key] = val
+        return new_params
+    if isinstance(params, list):
+        new_params = []
+        for val in params:
+            if isinstance(val, dict) or isinstance(val, list):
+                val = clean_params(val)
+            if val:
+                new_params.append(val)
+        return new_params
```
EarthmanT, line 124: Maybe you want to move this to utils.py?

cloudify_aws/ec2/resources/image.py
kaplanyaniv: image creation docent work well with wait for status decorator

cloudify_aws/ec2/resources/image.py
kaplanyaniv: image dergistrtion docent work well with wait for status decorator

cloudify_aws/ec2/resources/image.py
kaplanyaniv: minimized it

PR #421, opened 2022-01-23
store resou8rce in runtime props in poststart

EarthmanT:


kaplanyaniv (commented):
I think we should bump the version and update the changeling

PR #423, opened 2022-01-30
Rd 3922 use existing creates vpc - node group fix

kaplanyaniv:


cloudify_aws/ec2/resources/nat_gateway.py
```diff
@@ -141,7 +141,17 @@ def create(ctx, iface, resource_config, **_):
         params[ALLOCATION_ID] = allocation_id
 
     # Actually create the resource
-    create_response = iface.create(params)['NatGateway']
+    try:
+        create_response = iface.create(params)['NatGateway']
+    except ClientError as e:
+        if 'MissingParameter' in str(e):
+            raise RuntimeError('AWS create_nat_gateway api has changed. '
```
EarthmanT, line 148: We should actually make this a NonRecoverableError

cloudify_aws/eks/resources/node_group.py
```diff
@@ -44,16 +44,44 @@ class EKSNodeGroup(EKSBase):
...
+        self._describe_param = value
+
+    def wait_for_status(self):
+        self.logger.info('Performing wait for node group.')
+        try:
+            self.wait_for_nodegroup(
+                self.describe_params, 'nodegroup_active',
+                max_attempt=30)
+        except WaiterError:
+            raise OperationRetry('Waiting for nodegroup...')
```
EarthmanT, line 74: Maybe we should also add this to cluster.py since we know it uses the same thing?

EarthmanT (changes requested):

cloudify_aws/eks/resources/cluster.py
```diff
@@ -144,13 +143,16 @@ def wait_for_cluster(self, params, status):
...
+        try:
+            waiter.wait(
+                name=params.get(CLUSTER_NAME),
+                WaiterConfig={
+                    'Delay': 30,
+                    'MaxAttempts': 40
+                }
+            )
+        except WaiterError:
+            raise OperationRetry('Waiting for cluster...')
```
EarthmanT, line 156: I think we should add the `wait_for_status` thing to this class.

EarthmanT (changes requested):

PR #465, opened 2022-07-04
fix resource tags

EarthmanT:


v2_plugin.yaml
```diff
@@ -4300,6 +4300,6 @@ labels:
       - aws
 
 resource_tags:
-  tenant: get_sys [tenant, name]
-  deployment_id: get_sys [deplyment, id]
-  owner: get_sys [deplyment, owner]
+  tenant: { get_sys: [ tenant, name ] }
+  deployment_id: { get_sys: [ deplyment, id ] }
```
ahmadiesa-abu, line 4304:
```suggestion
  deployment_id: { get_sys: [ deployment, id ] }
```

ahmadiesa-abu (changes requested):
@EarthmanT I think we still need to bump the version to publish new version with fixes

ahmadiesa-abu (approved):

PR #513, opened 2023-03-24
RD-6874 Adding notify to slack feature.

bartoszkosciug:
Adding feature to notify slack about failed job + store tests results of unittests + store artifacts of other (custom) tests.


EarthmanT (approved):

yuvlar (approved):
Looks very good.
I do wonder if there's a way to natively add the slack-secrets context to the whole file instead of naming it in every workflow. but that's for the future

PR #541, opened 2023-05-23
RND-576-s-3-example-blueprint-fails-with-credentials

Nelynehemia:


cloudify_aws/s3/resources/bucket.py
```diff
@@ -110,6 +125,10 @@ def create(ctx, iface, resource_config, params, **_):
 
     # Actually create the resource
     bucket = iface.create(params)
```
EarthmanT, line 127:
```suggestion
    try:
        bucket = iface.create(params)
    except BaseException as e:
        if 'Bucket cannot have ACLs set with ObjectOwnership's BucketOwnerEnforced setting' in str(e):
            ctx.logger.error('Deprecation warning, the AWS API has changed and ACL-public.......is no longer valid.')
            acl = params.pop('ACL')
            if acl == 'public-read-write':
                iface.create(params)
                iface.put_public_access_block(params)
            else:
                raise e
        else:
            raise e
```

cloudify_aws/s3/resources/bucket.py
```diff
@@ -109,7 +123,25 @@ def create(ctx, iface, resource_config, params, **_):
             del params['CreateBucketConfiguration']
 
     # Actually create the resource
-    bucket = iface.create(params)
+    try:
+        bucket = iface.create(params)
+    except BaseException as e:
+        ctx.logger.error(str(e))
```
EarthmanT, line 129: I do not think that we need to log the exception.

We do need to check here that the error message indicating that ACL is the reason that this failed. For example
```
except BaseException as e:
   if 'the ACL is the problem' in str(e):
      # Then do the put_public_access_block thing.
```

cloudify_aws/s3/resources/bucket.py
```diff
@@ -109,7 +123,25 @@ def create(ctx, iface, resource_config, params, **_):
...
+            if 'public-read' in acl:
+                bucket = iface.create(params)
+                iface.put_public_access_block(params)
+            else:
+                raise e
+        else:
+            raise e
+
+    # PublicAccessBlockConfiguration
+    iface.put_public_access_block(params)
```
EarthmanT, line 143: This is already called in line 136. Why do we call it again?

cloudify_aws/s3/resources/bucket_object.py
```diff
@@ -247,7 +247,20 @@ def create(ctx, iface, resource_config, **_):
     ctx.instance.runtime_properties[BUCKET] = bucket_name
 
     # Actually create the resource
-    iface.create(resource_config)
+    try:
+        iface.create(resource_config)
+    except BaseException as e:
+        ctx.logger.error(str(e))
```
EarthmanT, line 253: I don't think that we need to log the exception.

cloudify_aws/s3/resources/bucket_object.py
```diff
@@ -247,7 +247,17 @@ def create(ctx, iface, resource_config, **_):
     ctx.instance.runtime_properties[BUCKET] = bucket_name
 
     # Actually create the resource
-    iface.create(resource_config)
+    try:
+        iface.create(resource_config)
+    except NonRecoverableError as e:
+        acl = resource_config.pop('ACL', '')
+        if 'AccessControlListNotSupported' not in str(e) \
+            and 'The bucket does not allow ACLs' not in str(e) \
```
EarthmanT, line 255: I don't thinks `and 'The bucket does not allow ACLs' not in str(e)` is necessary.
```suggestion
```

cloudify_aws/s3/resources/bucket_object.py
```diff
@@ -247,7 +247,17 @@ def create(ctx, iface, resource_config, **_):
...
+                and 'public-read' not in acl:
+            raise e
+        ctx.logger.error('Deprecation warning, the AWS API has changed and \
+                         ACL-public is no longer valid.')
+        iface.create(resource_config)
```
EarthmanT, line 261:
```suggestion
        iface.put_public_access_block(resource_config)
```

EarthmanT (approved):

PR #549, opened 2023-07-20
added utils function to validate aws id is good

kaplanyaniv:


cloudify_aws/ec2/resources/eni.py
```diff
@@ -298,6 +299,10 @@ def _create_eni_params(params, ctx_instance):
                 EXTERNAL_RESOURCE_ID)
         if group_id and group_id not in groups:
             groups.append(group_id)
+    for group_id in groups:
+        if not utils.is_valid_aws_id('sg', group_id):
+            # TODO add error message
+            raise NonRecoverableError()
```
EarthmanT, line 305:
```suggestion
            ctx.logger.error('Not adding provided security group to group list, because it is malformed ID'.format(group_id))
```

EarthmanT (approved):

